PR #20, opened 2025-06-09
11: Set up NestJS

### LINKED ISSUE #11: Set up NestJS projects
**AS A** Developer
**I WANT** to have a proper project setup
**SO THAT** I can develop effectively

## ❗ Quick shortcut

I have a [starter project](https://github.com/IgorKrupenja/nestjs-starter) ready, which has most of the stuff configured and will fulfil most of the acceptance criteria below. You can use it as a starting point. If you do, please remove things we will not need:

- [ ] Everything related to Prisma and Postgres, including dependencies.

## Acceptance criteria

- [x] Project should be configured to use latest Node LTS and latest pnpm, **not** npm. pnpm and Node versions should be enforced in `pacakge.json` and `.nvmrc`.
- [x] NestJS project created with the latest dependency versions.
- [x] Prettier set up with [rules like in the example project](https://github.com/IgorKrupenja/nestjs-starter/blob/main/.prettierrc).
- [x] ESLint set up with [rules like in the example project](https://github.com/IgorKrupenja/nestjs-starter/blob/main/eslint.config.mjs).
- [x] ❗ Tests are obligatory in this project. You are free to choose whether to use unit tests, e2e tests **or** a combination of both. Please create test configuration with Vitest.
  - [x] Sample unit test configuration is there in the [example project](https://github.com/IgorKrupenja/nestjs-starter).
  - [x] **Not in starter repo**: e2e tests can also be configured to use Vitest, please see [NestJS docs](https://docs.nestjs.com/recipes/swc#vitest).
- [x] Project has a short Readme, detailing how to run the project, tests, etc. — you can also modify the existing Readme from the example project.
- [x] Dockerfile is there and can be built with `docker build . --no-cache`
- [x] **Not in starter repo**: When you are done with configuring NestJS for one project, please copy it so that we have two directories with projects in the root of the repo: `dmr-server` and `dmr-agent`.
- [x] **Not in starter repo**: Dependabot [is configured](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference) in this repository for npm packages in **both** projects. It is configured to create PRs automatically.
- [x] **Not in starter repo**: Github Actions are configured for tests, lint, format and build. Github Actions should only be run on non-draft PRs. Example project has [a decent config](https://github.com/IgorKrupenja/nestjs-starter/tree/main/.github/workflows). However, since we have two Nest projects in this repo, `dmr-server` and `dmr-agent`, we would need some extra steps:
  - [x] Create two separate workflow files, e.g. `run-checks-dmr-agent.yaml` and `run-checks-dmr-server.yaml`. You can even create a common workflow file and just pass it the path in agent/server workflow files.
  - [x] Please make sure that only the relevant workflows (agent or server) are run depending on which directory had changes. [See this](https://stackoverflow.com/questions/63822219/how-to-run-github-actions-workflow-only-if-the-pushed-files-are-in-a-specific-fo).
- [ ] A shared project for common DTOs, etc is added. This common parts can be used by both `dmr-server` and `dmr-agent`. Please use your best judgement on how to implement this. Right now it should only contain one DTO/interface, name e.g. `AgentConfiguration` with the data structure (pesudo-code) below:

```
AgentConfiguration {
	id: string/UUID,
	name: string,
	authenticationCertificate: string, // this is the public key
	createdAt: string,
	updatedAt: string,
}
```

furqattrade:
Related task: https://github.com/buerokratt/DMR/issues/11


.github/workflows/run-check.yml
```diff
@@ -0,0 +1,29 @@
...
+        type: string
+
+jobs:
+  check:
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    defaults:
+      run:
+        working-directory: ${{ inputs.working-directory }}
```
IgorKrupenja, line 19: Thanks, this is clever!

.github/workflows/run-checks-dmr-client.yml
```diff
@@ -0,0 +1,50 @@
+name: Run checks (DMR Client)
```
IgorKrupenja: - [ ] Please rename this to `Run checks (DMR Agent)`. We would like to use this term throughout the project.
- [ ] Please also rename the file to `run-checks-dmr-agent.yml`

README.md
```diff
@@ -1,68 +1,65 @@
...
-- In the future, can potentially be extended to perform operations — like applying policies — on incoming and outgoing messages.
+https://koodivaramu.eesti.ee/buerokratt/architecture/concepts-and-proof-of-concepts/-/tree/master/Distributed%20Message%20Rooms
 
-### RabbitMQ
+### Related links
 
-- Has per-Agent message queues.
-- Has a dead letter queue for messages that failed to deliver.
-- Has RabbitMQ UI-based monitoring tools set up.
-- Supports RabbitMQ clustering for scalability.
```
IgorKrupenja, line 68: Please revert the changes to the Readme. The `dev` branch now has the updated Readme content, we really do not want to have the confusing outdated content there.

dmr-server/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+FROM base AS build
+
+WORKDIR /app
+COPY . .
+COPY --from=dependencies /app/node_modules ./node_modules
+RUN pnpm install --frozen-lockfile
+RUN pnpm run build
+
+FROM node:22.14-alpine AS deploy
```
IgorKrupenja: Please update the version to `22.16`. It was a typo in my starter project.

dmr-server/nest-cli.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "https://json.schemastore.org/nest-cli",
+  "collection": "@nestjs/schematics",
+  "sourceRoot": "src",
+  "compilerOptions": {
+    "deleteOutDir": true
```
IgorKrupenja: Just curious, any reasons not to use `"builder": "swc"` here? It is not obligatory but SWC is generally a bit faster for local development in my experience.

.github/workflows/run-checks-dmr-client.yml
```diff
@@ -0,0 +1,50 @@
...
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'dmr-client/**'
```
IgorKrupenja: Please do the following:

- [ ] Copy over the files from `dmr-server` directory to `dmr-agent` directory. We would like to have the Agent set up ASAP as well.
- [ ] Please use `dmr-agent` as a directory name (and **not** `dmr-client`). Please also change this in this YAML file everywhere. The task originally had a mistake in the naming, now fixed, sorry!

IgorKrupenja (changes requested):

.github/dependabot.yml
```diff
@@ -0,0 +1,53 @@
...
+    directory: "/dmr-server"
+    schedule:
+      interval: "daily"
+      time: "03:00"
+      timezone: "Asia/Tashkent"
+    labels:
+      - "dependencies"
+      - "dmr-server"
+    open-pull-requests-limit: 10
+    target-branch: "main"
```
IgorKrupenja: Please change the target branch to `dev` for now.

.github/dependabot.yml
```diff
@@ -0,0 +1,53 @@
...
+    directory: "/dmr-client"
+    schedule:
+      interval: "daily"
+      time: "03:00"
+      timezone: "Asia/Tashkent"
+    labels:
+      - "dependencies"
+      - "dmr-client"
+    open-pull-requests-limit: 10
+    target-branch: "main"
```
IgorKrupenja: Please change the target branch to `dev` for now.

.github/dependabot.yml
```diff
@@ -0,0 +1,53 @@
+version: 2
+updates:
+  # DMR Server dependencies
+  - package-ecosystem: "npm"
+    directory: "/dmr-server"
+    schedule:
+      interval: "daily"
+      time: "03:00"
+      timezone: "Asia/Tashkent"
```
IgorKrupenja: Please use Tallinn time zone. This will be more descriptive to the client.

.github/dependabot.yml
```diff
@@ -0,0 +1,53 @@
...
+          - "vitest"
+          - "@vitest/*"
+
+  # DMR Client dependencies
+  - package-ecosystem: "npm"
+    directory: "/dmr-client"
+    schedule:
+      interval: "daily"
+      time: "03:00"
+      timezone: "Asia/Tashkent"
```
IgorKrupenja: Please use Tallinn time zone. This will be more descriptive to the client.

.github/dependabot.yml
```diff
@@ -0,0 +1,53 @@
...
+    labels:
+      - "dependencies"
+      - "dmr-client"
```
IgorKrupenja: Please change this to `dmr-agent` as well.

.github/dependabot.yml
```diff
@@ -0,0 +1,53 @@
...
+        patterns:
+          - "@typescript-eslint/*"
+      vitest:
+        patterns:
+          - "vitest"
+          - "@vitest/*"
+
+  # DMR Client dependencies
+  - package-ecosystem: "npm"
+    directory: "/dmr-client"
```
IgorKrupenja: Please change this to `dmr-agent` as well.

IgorKrupenja (changes requested):

dmr-server/src/health/health.controller.ts
```diff
@@ -0,0 +1,13 @@
+import { Controller, Get } from '@nestjs/common';
+
+import { HealthService } from './health.service';
+
+@Controller('health')
```
IgorKrupenja: We would like the endpoints to be versioned just in case. Please change this line to:

```ts
@Controller('/v1/health')
```

IgorKrupenja (changes requested):

dmr-server/src/health/health.controller.ts
furqattrade: what about putting version in env and make it global for all controllers by prefix in main.ts

IgorKrupenja: Honestly, I do not like this. What happens if we need both old and new version at the same time? E.g. for some transition period. This has happened before with other projects. I do not think this is very likely with DMR but I would still keep the version prefix in controllers for now.

PR #23, opened 2025-06-10
18: Set up RabbitMQ 

### LINKED ISSUE #18: Set up RabbitMQ
**AS AN** Developer
**I WANT** to have a basic RabbitMQ node set up
**SO THAT** DMR server can use it for queueing messages

## Acceptance criteria

- [ ] `rabbitmq` folder in the repo root has a simple `docker-compose.yaml` file with one RabbitMQ node, similar to [this](https://medium.com/@kaloyanmanev/how-to-run-rabbitmq-in-docker-compose-e5baccc3e644).
- [ ] RabbitMQ must be configured in a way that queues and their data are persisted after node restart.
- [ ] RabbitMQ must be configured in a way that events from queues are moved to a dead letter queue after **configurable** TTL timeout.

aNazirov:


dmr-server/libs/rabbitmq/src/rabbitmq.service.ts
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable prettier/prettier */
```
IgorKrupenja: Why is this necessary? Is it causing issues? We should not be disabling linter rules in general.

dmr-server/libs/rabbitmq/tsconfig.lib.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "declaration": true,
+    "outDir": "../../dist/libs/rabbitmq"
+  },
+  "include": ["src/**/*"],
+  "exclude": ["node_modules", "dist", "test", "**/*spec.ts"]
+}
```
IgorKrupenja, line 9: Just curious: why would we want to have RabbitMQ related stuff in a separate `libs` dir and not just a as a regular module under `src`?

IgorKrupenja (changes requested):

dmr-server/libs/rabbitmq/src/rabbitmq.service.ts
aNazirov: Yes, of course. I had a few error messages like "Delete ␍", which is why I disabled Prettier for these files. I plan to fix it after we complete setting up our project.

dmr-server/libs/rabbitmq/tsconfig.lib.json
aNazirov: This is a recommendation from Nest.js: we expose clients for working with specific tools in the `libs` folder instead of keeping them in the business layer, since they are considered more of a tool rather than a module from Nest.js's perspective.

dmr-server/libs/rabbitmq/src/rabbitmq.service.ts
aNazirov: Fixed

dmr-server/libs/rabbitmq/tsconfig.lib.json
IgorKrupenja: Fine then, thanks!

dmr-server/libs/rabbitmq/src/rabbitmq.service.ts
IgorKrupenja: Thanks!

aNazirov:
Completed RabbitMQ client and finished setting up the config module.

dmr-server/libs/rabbitmq/src/rabbitmq.service.ts
```diff
@@ -0,0 +1,48 @@
...
+    await this.channel.assertQueue(queueName, {
+      durable: true,
+      arguments: {
+        'x-message-ttl': ttl ?? this.ttl, // 30 seconds
+        'x-dead-letter-exchange': '', // use default exchange
+        'x-dead-letter-routing-key': dlqName,
+      },
+    });
+
+    console.log(`Queue ${queueName} with TTL ${ttl ?? this.ttl}ms and DLQ ${dlqName} set up.`);
```
IgorKrupenja: Can we please use `this.logger.log` here as well?

dmr-server/eslint.config.mjs
```diff
@@ -76,6 +76,12 @@ export default tseslint.config(
           ignoreDeclarationSort: true,
         },
       ],
+      'prettier/prettier': [
+        'error',
+        {
+          endOfLine: 'auto',
```
IgorKrupenja, line 82: Aha! Great catch. I assume not having this was causing CR/LF linter errors before?

Thanks!

IgorKrupenja (changes requested):

dmr-server/eslint.config.mjs
aNazirov: Yep

dmr-server/libs/rabbitmq/src/rabbitmq.service.ts
aNazirov: Of course, my bad

aNazirov: Thank you for mentioning it, I did see that.

aNazirov: Done

IgorKrupenja:
Merging with broken GitHub Actions for now, these are resolved in another PR.

PR #24, opened 2025-06-10
Get configuration from CentOps #13

### LINKED ISSUE #13: Get configuration from CentOps
**AS AN** Architect
**I WANT** DMR Server to get configuration from CentOps
**SO THAT** I can use it to create queues in RabbitMQ and pass on the configuration to DMR Agents

## Acceptance criteria

- [ ] Task service is created to get the configuration from CentOps on schedule. Please see [this](https://docs.nestjs.com/techniques/task-scheduling#declarative-cron-jobs).
- [ ] Environment variable for the frequency of getting configuration is added. Name something like `CENTOPS_CONFIGURATION_CRON_TIME` or similar. Please set it for once every 1 hour in development `.env` file for now.
- [ ] Another env variable — `CENTOPS_CONFIGURATION_URL` name or similar — is added. You can set it to some dummy value ending with `/centops/clients` for now.
- [ ] CentOps configuration URL is polled periodically based on the env variable values created above. Since CentOps is only being developed, we would need to mock these requests for now. This mock are configured in away so that it can be removed with minimal code changes. One possible option to achieve this is to [use MSW](https://mswjs.io/docs/integrations/node/). You can enable the mock only locally based on `NODE_ENV` value.
- [ ] The mocked CentOps endpoint should return data in the following format:

```json
{
  "response": [
    {
      "id": "d3b07384-d9a0-4c3f-a4e2-123456789abc",
      "name": "Police",
      "authentication_certificate": "-----BEGIN CERTIFICATE-----\nMIID...==\n-----END CERTIFICATE-----", // This is the client public key
      "created_at": "2025-06-10T12:34:56Z",
      "updated_at": "2025-06-10T12:34:56Z"
    },
    {
      "id": "a1e45678-12bc-4ef0-9876-def123456789",
      "name": "Tax office",
      "authentication_certificate": "-----BEGIN CERTIFICATE-----\nABCD...==\n-----END CERTIFICATE-----",
      "created_at": "2025-06-08T08:22:10Z",
      "updated_at": "2025-06-09T09:13:44Z"
    }
  ]
}
```

- [ ] The configuration data needs to be stored in-memory only and needs to be accessible in every module of DMR Server. Please research this but [caching suggested by NestJS](https://docs.nestjs.com/techniques/caching) is probably good enough. Or even easier, use a private variable in some service. Choose whichever makes more sense and seems easier to do. Only the following data is stored: `id`, `name`, `authenticationCertificate`, `createdAt`, `updatedAt`. Please use a common DTO for this, implemented [in this task](https://github.com/buerokratt/DMR/issues/11). The data is fully overwritten on each response from CentOps.
- [ ] Tests to cover this functionality are added.

## Not in scope

- Authentication between DMR Server and CentOps, there is going to be a separate task for this.

sherifbai:
Related task: https://github.com/buerokratt/DMR/issues/13


dmr-server/src/main.ts
```diff
@@ -10,6 +10,15 @@ async function bootstrap(): Promise<void> {
...
   app.useGlobalPipes(new ValidationPipe({ whitelist: true, forbidNonWhitelisted: true }));
 
+  if (process.env.NODE_ENV !== 'production') {
+    const { setupServer } = await import('msw/node');
+    const { handlers } = await import('./mocks/handlers/response');
+
+    const server = setupServer(...handlers);
+    server.listen();
+    console.log('MSW server started for development/testing.');
+  }
```
IgorKrupenja: We already have an `if (process.env.NODE_ENV === 'development')` block below. Could you please move this code there? This will be more clear. Also, you can then use `logger.log` instead of `console.log`.

dmr-server/src/common/config/app.config.ts
```diff
@@ -0,0 +1,17 @@
...
+  (): CentOpsConfig => ({
+    url: process.env.CENTOPS_CONFIGURATION_URL || '',
+    cronTime: process.env.CENTOPS_CONFIGURATION_CRON_TIME || '*/30 * * * *',
+  }),
+);
+
+export type CentOpsConfig = {
+  cronTime: string;
+  url: string;
+};
```
IgorKrupenja, line 18: Is there a reason not to use a separate [config module as recommended by NestJS](https://docs.nestjs.com/techniques/configuration)? Also, Bobur mentioned yesterday that he expects there to be a config module.

dmr-server/src/modules/centops/dto/client-config.dto.ts
```diff
@@ -0,0 +1,34 @@
...
+  @IsString()
+  @IsNotEmpty()
+  createdAt: string;
+
+  @IsString()
+  @IsNotEmpty()
+  updatedAt: string;
+}
+
+export class CentOpsClientResponseDto {
```
IgorKrupenja: Could you please extract this DTO into its own file? Also, can this be just an `interface` since we are not validating anything?

IgorKrupenja (changes requested):
Extra:

- [ ] Please rename the PR to have the GitHub issue ID in the name.
- [ ] Please add the GitHub issue link to the PR description.
- [ ] GitHub Actions are failing because `packageManager` is missing in `package.json`. I missed this before, my bad. Could you please add this single line to the root of `package.json`:

```json
"packageManager": "pnpm@10.12.1",
```

dmr-server/src/common/config/app.config.ts
aNazirov: There seems to be a slight misunderstanding due to the naming, but everything was done following Nest’s recommendations. If possible, you can skip this point — I’ll adjust the naming and finish configuring it on my end.

IgorKrupenja: Yes, of course! If this is fine by you, let it be. I'll resolve this conversation.

aNazirov: Thank you)

sherifbai:
image
there is already packageManager

IgorKrupenja:
> image there is already packageManager

Yes, good point. This is strange but to move things forward, let's merge for now. The actions are working fine on Furqat's branch with improved repo structure so this should be fixed when we merge that one.

PR #28, opened 2025-06-11
21, 33: Managing agent queues in RabbitMQ and configure DLQ TTL

### LINKED ISSUE #21: Managing agent queues in RabbitMQ
**AS AN** Architect
**I WANT** DMR Server to manage RabbitMQ queues
**SO THAT** I can create and delete queues based on the configuration from CentOps

## Acceptance criteria

- [ ] On getting configuration from CentOps (implemented in https://github.com/buerokratt/DMR/issues/13), DMR server compares the list of agents it has in memory to the new list, by `id` field.
- [ ] If it sees a new agent:
	- [ ] It should create a new queue in RabbitMQ. The ID of the queue should be set to `id` of the agent from CentOps data. 
	- [ ] Also, a DLQ (dead letter queue) for that agent should be created, see [this discussion with Bobur on Slack](https://devtailor.slack.com/archives/C0918SMVCLQ/p1749546202962229).
	- [ ] DLQ should be configured in a way that messages from normal queues are moved there after a configurable (via env variable) TTL timeout. Set the default of env variable to 5 minutes for now. The timeout should be the same for all agent queues.
- [ ] If it see that an agent has been removed in the CentOps config, it should delete the normal queue and DLQ for that agent (by ID).

## Notes

- Please feel free to implement this according to your best judgement, but maybe [this](https://www.freecodecamp.org/news/message-queues-with-rabbitmq-in-nest-js/) has some useful code bits.

### LINKED ISSUE #33: Configure DLQ TTL
**AS AN** Architect
**I WANT** DLQs to have a TTL
**SO THAT** we also purge DLQ messages after a certain period

- [ ] A new env variable is added to configure TTL for DLQs. This is separate from the TTL defined as `RABBITMQ_DEFAULT_TTL` [here](https://github.com/buerokratt/DMR/pull/28/files#diff-6857f9435b48108708e28096e94166724f9e16fa3a6beed821a5bb05a5f16fe0R6).
- [ ] This variable is passed as a TTL when creating **DLQs**, not normal queues.
- [ ] The default value is 24 hours.
- [ ] Please check this in a test if not too difficult.

aNazirov:
Related task: https://github.com/buerokratt/DMR/issues/21


dmr-server/src/modules/centops/centops.service.ts
```diff
@@ -5,49 +5,62 @@ import { SchedulerRegistry } from '@nestjs/schedule';
...
     private readonly schedulerRegistry: SchedulerRegistry,
     @Inject(CACHE_MANAGER) private cacheManager: Cache,
   ) {}
 
   onModuleInit(): void {
-    const job = new CronJob(this.centOpsConfig.cronTime, async (): Promise<void> => {
+    const onTick = async (): Promise<void> => {
       this.logger.debug(
-        `Executing cron job '${this.CENTOPS_JOB_NAME}' at ${new Date().toISOString()}`,
+        `Executing cron job '${this.CENT_OPS_JOB_NAME.toString()}' at ${new Date().toISOString()}`,
```
IgorKrupenja: Why do we need `.toString()` here? `CENT_OPS_JOB_NAME` is a string already.

dmr-server/src/modules/centops/centops.service.ts
```diff
@@ -5,49 +5,62 @@ import { SchedulerRegistry } from '@nestjs/schedule';
...
+
+    const job = new CronJob(this.centOpsConfig.cronTime, onTick);
+
+    this.schedulerRegistry.addCronJob(this.CENT_OPS_JOB_NAME, job);
+
     job.start();
+
     this.logger.log(
-      `Cron job '${this.CENTOPS_JOB_NAME}' scheduled for: ${this.centOpsConfig.cronTime}`,
+      `Cron job '${this.CENT_OPS_JOB_NAME.toString()}' scheduled for: ${this.centOpsConfig.cronTime}`,
```
IgorKrupenja: Why do we need `.toString()` here? `CENT_OPS_JOB_NAME` is a string already.

dmr-server/libs/rabbitmq/src/rabbitmq.service.ts
```diff
@@ -37,12 +37,28 @@ export class RabbitMQService implements OnModuleInit {
...
+
+    // Delete our queue
+    await this.channel.deleteQueue(queueName);
+
+    this.logger.log(`Queue ${queueName} and DLQ ${dlqName} deleted.`);
+  }
+
+  private getDLQName(queueName: string): string {
+    return `${queueName}_dlq`;
+  }
```
IgorKrupenja: Would it be possible to add tests for this service already?

dmr-server/libs/rabbitmq/src/rabbitmq.service.ts
```diff
@@ -37,12 +37,28 @@ export class RabbitMQService implements OnModuleInit {
     await this.channel.assertQueue(queueName, {
       durable: true,
       arguments: {
-        'x-message-ttl': ttl ?? this.ttl, // 30 seconds
+        'x-message-ttl': ttl ?? this.ttl,
```
IgorKrupenja: Please make this TTL configurable via an env variable. Set any value you like in `.env.example`, e.g. 5 minutes.

IgorKrupenja (changes requested):

dmr-server/src/modules/centops/centops.service.ts
aNazirov: I wanted to replace the string with a Symbol but ran into a tool limitation and forgot to revert that change.

dmr-server/src/modules/centops/centops.service.ts
aNazirov: I wanted to replace the string with a Symbol but ran into a tool limitation and forgot to revert that change.

dmr-server/libs/rabbitmq/src/rabbitmq.service.ts
aNazirov: Got it

dmr-server/libs/rabbitmq/src/rabbitmq.service.ts
aNazirov: Yes of course, covered the RabbitMQ service with tests.

dmr-server/libs/rabbitmq/src/rabbitmq.service.ts
aNazirov: Done

dmr-server/src/modules/centops/centops.service.ts
aNazirov: Fixed

dmr-server/src/modules/centops/centops.service.ts
aNazirov: Fixed

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
```diff
@@ -0,0 +1,85 @@
...
+      // Create DLQ for our queue
+      await this.channel.assertQueue(dlqName, { durable: true });
+
+      // Create and setup our queue
+      await this.channel.assertQueue(queueName, {
+        durable: true,
+        arguments: {
+          'x-message-ttl': ttl ?? this.rabbitMQConfig.ttl,
+          'x-dead-letter-exchange': '', // use default exchange
+          'x-dead-letter-routing-key': dlqName,
```
IgorKrupenja, line 46: - [ ] As discussed on Slack, please also set `x-queue-type` to `quorum` here, [more context](https://www.rabbitmq.com/docs/quorum-queues#declaring).
- [ ] If not very difficult, please also modify the tests to check for this.

apps/dmr-agent/vitest.config.ts
```diff
@@ -1,15 +1,19 @@
...
-    include: ['src/**/*.spec.ts'],
+    include: ['src/**/*.spec.ts', 'libs/**/*.spec.ts'],
     environment: 'node',
     root: './',
   },
   resolve: {
     alias: {
       '@src': './src',
+      'libs/rabbitmq': path.resolve(__dirname, './libs/rabbitmq/src'),
+      'libs/rabbitmq/*': path.resolve(__dirname, './libs/rabbitmq/src/*'),
```
IgorKrupenja: I do not think I get this. Why do we need to change anything in `dmr-agent`?

apps/dmr-server/src/modules/centops/centops.service.spec.ts
```diff
@@ -1,16 +1,18 @@
...
+
 import { CentOpsService } from './centops.service';
 import { centOpsConfig } from '../../common/config';
+import { RabbitMQService } from '../../libs/rabbitmq';
 
 describe('CentOpsService', () => {
   let service: CentOpsService;
   let httpService: HttpService;
-  let cacheManager: any;
+  let cacheManager: Cache;
```
IgorKrupenja, line 15: Great, thanks! 😊

package.json
```diff
@@ -39,7 +39,7 @@
     "class-transformer": "^0.5.1",
     "class-validator": "^0.14.2",
     "compression": "^1.8.0",
-    "cron": "^4.3.1",
+    "cron": "4.3.0",
```
IgorKrupenja, line 43: Why do we need this specific cron version?

IgorKrupenja (changes requested):
Extra:

- [ ] Please check why some of the GitHub Actions are failing.

apps/dmr-agent/vitest.config.ts
aNazirov: I didn't touch to dmr-agent, I think it happened during the merge when setting up Nx. I'll fix it

package.json
aNazirov: It needs to fix the TypeScript/ESLint error, as Furqat mentioned yesterday. The `@nestjs/schedule` package requires a specific version of `cron`, so if we continue using a different version, it will throw an error.

IgorKrupenja: Great, thanks!

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
aNazirov: About the second point - right now we’re only writing unit tests, but to verify quorum logic, we’d need to create integration tests. Unfortunately, we can’t do that at this stage.

aNazirov:
> Extra:
> 
> * [ ]  Please check why some of the GitHub Actions are failing.

There was an error with the test configuration, I fixed it

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
aNazirov: Upd: If you meant checking whether the x-queue-type argument was added - it’s done.

IgorKrupenja: Yes, this is what I meant, thanks!

IgorKrupenja (approved):

PR #35, opened 2025-06-12
26: Connecting to DMR Server with Socket.io

### LINKED ISSUE #26: Connecting to DMR Server with Socket.io
**AS AN** Architect
**I WANT** DMR Agents to be able to connect to DMR Server
**SO THAT** I can send messages and other events

## Related tasks

- DMR Server task: https://github.com/buerokratt/DMR/issues/15

## Acceptance criteria

- [ ] DMR Agent can establish a Socket.io connection with the DMR Server. DMR Server WebSocket URL is configurable via an env variable.
- [ ] On establishing the connection, the Agent should send a JWT token to perform authentication. The token is signed with Agent's **private key**, JWT payload has the following format:
```json
 {
   "sub": "cc5b7b04-ba33-4423-ba26-ccc25441db42", // This is the Agent ID
   "iat": 1717734000,
   "exp": 1717734300
 }
```
- [ ] `kid` header claim is set in JWT **header**, see [this](https://mojoauth.com/blog/what-is-jwt-key-id-kid/) or [this](https://stackoverflow.com/questions/43867440/whats-the-meaning-of-the-kid-claim-in-a-jwt-token). Its value is the Agent ID.
- [ ] JWT expiry time is set to 1 minute.
- [ ] Both Agent private key and Agent ID are configurable and passed as env variables. For local development, please use a generated private key matching the public key created in https://github.com/buerokratt/DMR/issues/15
- [ ] Socket.io suggested way of sending/receiving an auth token [here in docs](https://socket.io/docs/v4/client-options/#auth).
- [ ] The public-key cryptosystem is standard RSA so we can use `@nest/jwt` with `RS256` algorithm.
- [ ] Reconnection logic is implemented: needed in case connection to DMR server is lost or in case DMR Server did not get the new agent data from CentOps and auth fails. Should be very simple, similar to [this in docs](https://socket.io/docs/v4/connection-state-recovery). `reconnectionDelay` and `reconnectionDelayMax` are configurable with env variables.
- [ ] Some basic logging is added: on connection, error and reconnection, [docs](https://socket.io/docs/v4/client-socket-instance/#Events).

furqattrade:
https://github.com/buerokratt/DMR/issues/26


apps/dmr-agent/.env.example
```diff
@@ -0,0 +1,10 @@
+PORT=5001
+NODE_ENV='development'
+DMR_SERVER_WEBSOCKET_URL=''
+WEBSOCKET_DELAY_MAX=5000
+WEBSOCKET_RECONNECTION_DELAY=1000
+AGENT_PRIVATE_KET=''
```
IgorKrupenja: Please fix a typo, should be `AGENT_PRIVATE_KEY`, with **Y**.

apps/dmr-agent/src/modules/websocket/websocket.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  constructor(
+    private readonly configService: ConfigService,
+    private readonly jwtService: JwtService,
+  ) {}
+
+  async onModuleInit(): Promise<void> {
+    await this.connectToServer();
+  }
+
+  // eslint-disable-next-line @typescript-eslint/require-await
```
IgorKrupenja: Can we remove this and simply do this?

```ts
onModuleInit(): void {
  this.connectToServer();
}

private connectToServer(): void {
  ...
}
```

apps/dmr-agent/src/modules/websocket/websocket.service.ts
```diff
@@ -0,0 +1,149 @@
...
+    };
+
+    for (const [key, value] of Object.entries(requiredConfigs)) {
+      if (!value) {
+        const configName =
+          key === 'serverUrl'
+            ? 'DMR_SERVER_WEBSOCKET_URL'
+            : key === 'agentId'
+              ? 'AGENT_ID'
+              : 'AGENT_PRIVATE_KEY';
```
IgorKrupenja: This nested ternary is hard to read. Can we have something more readable? Simple if statements or some sort or a something like this:

```ts
const configNameMap = {
  serverUrl: "DMR_SERVER_WEBSOCKET_URL",
  agentId: "AGENT_ID",
  privateKey: "AGENT_PRIVATE_KEY",
} as const;

for (const [key, value] of Object.entries(requiredConfigs)) {
  if (!value) {
    const configName = configNameMap[key as keyof typeof configNameMap];
    this.logger.error(`${configName} is not configured`);
    return;
  }
}
```

apps/dmr-agent/src/modules/websocket/websocket.service.ts
```diff
@@ -0,0 +1,149 @@
+import { JwtPayload, SocketClient } from '@dmr/shared';
+import { Injectable, Logger, OnModuleInit } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import { JwtService } from '@nestjs/jwt';
+import { io } from 'socket.io-client';
+
+@Injectable()
+export class WebsocketService implements OnModuleInit {
+  private readonly logger = new Logger(WebsocketService.name);
+  private socket: SocketClient | null = null;
```
IgorKrupenja: `socket.io` has a built-in type `Socket`. 

- [ ] Please import it above and use here.
- [ ] Please remove the custom `SocketClient` type completely.

apps/dmr-agent/src/modules/websocket/websocket.service.ts
```diff
@@ -0,0 +1,149 @@
...
+        reconnectionDelay: this.configService.get<number>('WEBSOCKET_RECONNECTION_DELAY', 1000),
+        reconnectionDelayMax: this.configService.get<number>('WEBSOCKET_DELAY_MAX', 5000),
+        auth: (callback: (data: { token: string }) => void): void => {
+          callback({ token: this.generateJwtToken(agentId, privateKey) });
+        },
+        // @ts-expect-error - Socket.io types don't include connectionStateRecovery yet
+        connectionStateRecovery: {
+          maxDisconnectionDuration: 2 * 60 * 1000,
+        },
+      }) as unknown as SocketClient;
```
IgorKrupenja: `as unknown as SocketClient` cast should not be necessary if you use the `socket.io` built-in type `Socket`, see above. Please remove the cast.

apps/dmr-agent/src/modules/websocket/websocket.service.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    try {
+      const { serverUrl, agentId, privateKey } = requiredConfigs;
+      this.socket = io(serverUrl, {
+        reconnectionDelay: this.configService.get<number>('WEBSOCKET_RECONNECTION_DELAY', 1000),
+        reconnectionDelayMax: this.configService.get<number>('WEBSOCKET_DELAY_MAX', 5000),
+        auth: (callback: (data: { token: string }) => void): void => {
+          callback({ token: this.generateJwtToken(agentId, privateKey) });
+        },
```
IgorKrupenja: Is there a reason we need this callback here? Can't we simply do this?

```ts
auth: {
  token: this.generateJwtToken(agentId, privateKey),
},
```

apps/dmr-agent/src/modules/websocket/websocket.service.ts
```diff
@@ -0,0 +1,149 @@
...
+        // @ts-expect-error - Socket.io types don't include connectionStateRecovery yet
+        connectionStateRecovery: {
+          maxDisconnectionDuration: 2 * 60 * 1000,
+        },
```
IgorKrupenja: I do not think I get this. The docs [say](https://socket.io/docs/v4/connection-state-recovery#usage) that this is something configured on the server side. So this should be a part of DMR Server-side config for socket.io, right?

apps/dmr-agent/src/modules/websocket/websocket.service.ts
```diff
@@ -0,0 +1,149 @@
...
+      }) as unknown as SocketClient;
+
+      this.setupSocketEventListeners();
+    } catch (error_) {
```
IgorKrupenja: Please use just `error` without underscore.

apps/dmr-agent/src/modules/websocket/websocket.service.ts
```diff
@@ -0,0 +1,149 @@
...
+    });
+
+    this.socket.on('connect_error', (error: unknown) => {
+      this.reconnectionAttempts++;
+      const errorMessage =
+        error instanceof Error
+          ? error.message
+          : typeof error === 'string'
+            ? error
+            : 'Unknown error';
```
IgorKrupenja: Again, this nested ternary is a bit difficult to read. Honestly, I would do something similar to what you are already doing above:

```ts
const errorMessage = error instanceof Error ? error.message : String(error);
```

apps/dmr-agent/src/modules/websocket/websocket.service.ts
```diff
@@ -0,0 +1,149 @@
...
+      }
+    });
+
+    this.socket.on('reconnect_error', (error: unknown) => {
+      const errorMessage =
+        error instanceof Error
+          ? error.message
+          : typeof error === 'string'
+            ? error
+            : 'Unknown error';
```
IgorKrupenja: Again, this nested ternary is a bit difficult to read. Honestly, I would do something similar to what you are already doing above:

```ts
const errorMessage = error instanceof Error ? error.message : String(error);
```

apps/dmr-agent/src/modules/websocket/websocket.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  isConnected(): boolean {
+    return Boolean(this.socket?.connected);
+  }
+
+  disconnect(): void {
+    if (this.socket) {
+      this.socket.disconnect();
+      this.logger.log('Manually disconnected from DMR server');
+    }
+  }
```
IgorKrupenja, line 141: - [ ] Please also add a lifecycle method to run `this.socket.disconnect()` on destroy. We need this because in a Kubernetes environment, the pods may be restarted/killed and we want to disconnect cleanly. I forgot to mention this in acceptance criteria 😄  Something like this:

```ts
onModuleDestroy() {
   disconnect(); // or this.socket.disconnect() or similar
}
```

- [ ] If you decide to use the `disconnect` method you already have, please remove the word "Manually" from `logger.log` text.
- [ ] If not difficult, please add a unit test to check that `onModuleDestroy` calls disconnect.

apps/dmr-server/.env.example
```diff
@@ -0,0 +1,10 @@
+PORT=5000
+ENVIRONMENT=development
+
+CENTOPS_CONFIGURATION_URL = ''
+CENTOPS_CONFIGURATION_CRON_TIME = 'string'
+
+RABBITMQ_DEFAULT_PORT=' '
+RABBITMQ_DEFAULT_HOST=' '
+RABBITMQ_DEFAULT_USER=' '
+RABBITMQ_DEFAULT_PASS=' '
```
IgorKrupenja: If you add this file, could you please use more realistic values for env variables? The ones we would use when running locally.

apps/dmr-agent/.env.example
```diff
@@ -0,0 +1,10 @@
...
+AGENT_ID=''
```
IgorKrupenja: Please add some random UUID as an example agent ID here.

IgorKrupenja (changes requested):

apps/dmr-agent/.env.example
```diff
@@ -0,0 +1,12 @@
+PORT=5001
+NODE_ENV=development
+DMR_SERVER_WEBSOCKET_URL='http://localhost:5000'
+WEBSOCKET_DELAY_MAX=5000
+WEBSOCKET_RECONNECTION_DELAY=1000
+AGENT_PRIVATE_KEY='-----BEGIN RSA PRIVATE KEY-----
```
IgorKrupenja, line 7: Please add a comment above saying "Fake dummy key" in case anyone gets a wrong idea 😅

apps/dmr-agent/src/modules/websocket/websocket.service.ts
```diff
@@ -0,0 +1,142 @@
...
+      if (!value) {
+        const configName = configNameMap[key as keyof typeof configNameMap];
+        this.logger.error(`${configName} is not configured`);
+        return;
+      }
+    }
+
+    try {
+      const { serverUrl, agentId, privateKey } = requiredConfigs;
+      const socketOptions: Partial<ManagerOptions & SocketOptions> = {
```
IgorKrupenja, line 45: Thanks for adding proper types here, this is good!

apps/dmr-agent/src/modules/websocket/websocket.service.ts
IgorKrupenja: Thank you for removing this. I asked Sherifbai to add this config on the server.

IgorKrupenja (changes requested):

PR #36, opened 2025-06-12
15: Socket.io interface for DMR agents

### LINKED ISSUE #15: Socket.io interface for DMR agents
**AS AN** Architect
**I WANT** DMR Server to be able to accept socket.io connections
**SO THAT** I can connect from DMR Agents and send messages

## Related tasks

- DMR Agent task: https://github.com/buerokratt/DMR/issues/26

### General

- [ ] WebSocket gateway is created in DMR Server. Should largely follow [NestJS docs](https://docs.nestjs.com/websockets/gateways) but feel free to adjust as necessary. Please actually use **socket.io** for communication over WebSocket. We would prefer to use it due to built-in retry logic.
- [ ] The gateway should support accepting connections from several DMR agents at the same time.

### Auth

- [ ] When DMR agent first connects via the WebSocket, it should be authenticated by the DMR server. On establishing a connection, the Agent will send a JWT signed by that agent's **private key**, JWT payload will have the following format:

```json
{
  "sub": "cc5b7b04-ba33-4423-ba26-ccc25441db42", // This is the Agent ID
  "iat": 1717734000,
  "exp": 1717734300
}
```

- [ ] `kid` header claim will be set in JWT **header**, see [this](https://mojoauth.com/blog/what-is-jwt-key-id-kid/) or [this](https://stackoverflow.com/questions/43867440/whats-the-meaning-of-the-kid-claim-in-a-jwt-token). Its value is also the Agent ID.
- [ ] DMR server should get the **public key** for the Agent ID (from decoded header, `kid`) from it's memory cache and verify the the signature with that **public key** (`authenticationCertificate`).
- [ ] DMR server should verify that `exp` time has not passed.
- [ ] DMR server should verify that `kid` value matches the `sub` value
- [ ] Socket.io suggested way of sending/receiving an auth token [here in docs](https://socket.io/docs/v4/client-options/#auth)
- [ ] But please try to implement this in a way that we are using [standard NestJS guards](https://docs.nestjs.com/websockets/guards).
- [ ] The public-key cryptosystem is standard RSA so we can use `@nest/jwt` with `RS256` algorithm.
- [ ] A couple public-private key pairs we can use for local development are created. Mocked CentOps response uses those public keys.

### Consuming queue messages

- [ ] If authentication is successful, the DMR server should start consuming messages in the RabbitMQ queue for that agent. The queue ID is the DMR Agent ID it should get from `sub` in JWT.
	- [ ] If this task is started before https://github.com/buerokratt/DMR/issues/21 is ready, you can just leave a `TODO` comment in code where we should start consuming the messages from RabbitMQ. We are creating the queues in that other task. 
- [ ] When DMR agent disconnects from WebSocket, the DMR server should stop consuming messages from the the RabbitMQ queue for that agent.
	- [ ] If this task is started before https://github.com/buerokratt/DMR/issues/21 is ready, you can just leave a `TODO` comment in code where we should stop consuming the messages from RabbitMQ.

## Notes

- All other logic related to messages/events will be in other tasks.

sherifbai:
Related task: https://github.com/buerokratt/DMR/issues/15


apps/dmr-server/src/modules/auth/interfaces/heades.interface.ts
IgorKrupenja, line 1: Please fix typo in the file name, should be "heade**r**s".

apps/dmr-server/src/modules/centops/centops.service.ts
```diff
@@ -46,6 +46,19 @@ export class CentOpsService implements OnModuleInit {
...
   }
 
+  async getCentOpsConfigurationByClientId(clientId: string): Promise<ClientConfigDto> {
+    const centOpsConfigs = await this.cacheManager.get<ClientConfigDto[]>(
+      this.CENT_OPS_CONFIG_CACHE_KEY,
+    );
+
+    const clientConfig = centOpsConfigs.find((config) => config.id === clientId);
+    if (!clientConfig) {
+      throw new BadRequestException('Client configuration not found');
```
IgorKrupenja, line 63: In addition to throwing (this should correctly result in a 400 response), can you also use `this.logger.error` to log an error including the wrong `clientId`? This will be useful for debug/audit purposes.

apps/dmr-server/src/modules/auth/guards/ws-auth.guard.ts
```diff
@@ -0,0 +1,31 @@
...
+      client.handshake?.headers?.authorization?.replace('Bearer ', '')) as string;
+
+    if (!token) {
+      client.disconnect();
+
+      return false;
+    }
+
+    try {
+      const verify = await this.authService.verifyToken(token);
```
IgorKrupenja: Can you please use a slightly more descriptive name instead of `verify`? Maybe `verifiedToken`?

apps/dmr-server/src/modules/gateway/dmr.server.gateway.ts
```diff
@@ -0,0 +1,51 @@
...
+      client.disconnect();
+      return;
+    }
+
+    try {
+      const verify = await this.authService.verifyToken(token);
+      Object.assign(client, { user: verify });
+    } catch {
+      client.disconnect();
+    }
```
IgorKrupenja: I do not think I get this. Why do we need to run exactly the same auth logic as the one we already have in `WsAuthGuard.canActivate`? That should be enough, right?

apps/dmr-server/src/modules/gateway/types/types.d.ts
```diff
@@ -0,0 +1,7 @@
+import { IJwtPayload } from '@dmr/shared';
+
+declare module 'socket.io' {
+  interface Socket {
+    user: IJwtPayload;
```
IgorKrupenja: Does it have to be named `user`? If not, please use `agent` or `agentToken` or similar. This is somewhat confusing.

libs/shared/src/interfaces/jwt-payload.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface IJwtPayload {
```
IgorKrupenja: Furqat uses `JwtPayload` in [his PR](https://github.com/buerokratt/DMR/pull/35/files#diff-ee1a23c080eb537f925aaa80f19e7cd6a5199b7ec0a679de7a5d036b4e766c1fR1), without `I`. Let's use that to be consistent and avoid conflicts.

apps/dmr-server/src/modules/auth/interfaces/heades.interface.ts
```diff
@@ -0,0 +1,12 @@
+export interface IJwtHeader {
+  alg?: string;
+  typ?: string;
+  kid?: string;
+  [key: string]: unknown;
+}
+
+export interface IDecodedJwt {
```
IgorKrupenja: Let's use interface names without capital `I`, just `JwtHeader` etc. I missed this before, sorry! This way we stay consistent in our naming.

apps/dmr-server/src/modules/gateway/dmr.server.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+    loggerSpy.mockRestore();
+    vi.restoreAllMocks();
+  });
+
+  it('should be defined', () => {
+    expect(gateway).toBeDefined();
+  });
+
+  describe('handleConnection', () => {
+    const mockPayload = { sub: 'testAgentId', email: 'test@example.com' };
```
IgorKrupenja: Let's use a more realistic mock payload, remove `email`.

apps/dmr-server/src/modules/auth/auth.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  constructor(
+    private readonly centOpsService: CentOpsService,
+    private readonly jwtService: JwtService,
+  ) {}
+
+  async verifyToken(token: string) {
+    const clientId = this.getKidFromToken(token);
+    const clientConfig = await this.centOpsService.getCentOpsConfigurationByClientId(clientId);
+
+    return this.jwtService.verifyAsync<IJwtPayload>(token, {
```
IgorKrupenja: After verifying the token, we would also like to check the following: `sub` from payload should match `kid` from header **and** `exp` time from header has not yet passed.

- [ ] Can this be implemented here? If so, please also log errors if these checks fail.
- [ ] Please add tests for this too, if possible.

apps/dmr-server/src/modules/gateway/dmr.server.gateway.ts
```diff
@@ -0,0 +1,51 @@
...
+  OnGatewayDisconnect,
+  OnGatewayConnection,
+} from '@nestjs/websockets';
+import { Logger, UseGuards } from '@nestjs/common';
+import { Server, Socket } from 'socket.io';
+import { AuthService } from '../auth/auth.service';
+import { WsAuthGuard } from '../auth/guards/ws-auth.guard';
+
+@UseGuards(WsAuthGuard)
+@WebSocketGateway()
```
IgorKrupenja: - [ ] Please also configure `connectionStateRecovery`, I missed this. [Relevant doc](https://socket.io/docs/v4/connection-state-recovery#usage). I think something like this should work with NestJS:

```ts
@WebSocketGateway({
  connectionStateRecovery: {
    maxDisconnectionDuration: SOME_ENV_VAR,
    skipMiddlewares: true,
  },
})
```

- [ ] Make `maxDisconnectionDuration` configurable via an env variable. Set default value to 2 minutes.

apps/dmr-server/src/modules/gateway/dmr.server.gateway.ts
```diff
@@ -0,0 +1,51 @@
...
+export class DmrServerGateway implements OnGatewayConnection, OnGatewayDisconnect {
```
IgorKrupenja: Also, can you please rename this to just `AgentGateway`? and then the files too.

IgorKrupenja (changes requested):
Extra:

- [ ] Please also implement acceptance criteria from "Consuming queue messages" in the task

apps/dmr-server/src/modules/gateway/dmr.server.gateway.ts
sherifbai: Because auth guard will work only for @SubscribeMessage, it does not work for OnGatewayConnection or OnGatewayDisconnect interfaces

apps/dmr-server/src/modules/auth/auth.service.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+
+      throw new UnauthorizedException('Invalid token');
+    }
+
+    if (verifiedToken.sub !== clientId) {
+      this.logger.error('Token sub and kid do not match');
+
+      throw new BadRequestException('Token sub and kid do not match');
+    }
```
IgorKrupenja, line 42: What about `exp` verification? Or is this checked automatically with `jwtService.verifyAsync`? Can we have a test for this?

apps/dmr-server/src/modules/gateway/dmr.server.gateway.ts
IgorKrupenja: Yeah, this indeed seems to be the case and is rather unfortunate. In this case, I would actually remove the guard completely. I see no point in running this verification on every message to be honest, performing auth once while agent connects should be good enough.

apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -0,0 +1,63 @@
...
+  ) {}
+
+  async handleConnection(@ConnectedSocket() client: Socket): Promise<void> {
+    try {
+      const token: string = (client.handshake?.auth?.token ||
+        client.handshake?.headers?.authorization?.replace('Bearer ', '')) as string;
+
+      const jwtPayload = await this.authService.verifyToken(token);
+
+      void this.rabbitService.setupQueue(jwtPayload.sub);
```
IgorKrupenja: Perhaps my explanation in the task was not clear enough 😅 

We should not be creating/deleting the queues when agents connect or disconnect (by calling `setupQueue` and `deleteQueue`). The queues are created or deleted on getting changes from CentOps, this is already implemented.

Instead, what I meant was for the server to start and stop consuming when when agents connect or disconnect. So basically:

- [ ] On connect, we need to call [consume](https://amqp-node.github.io/amqplib/channel_api.html#channel_consume) for the queue of that agent. The queue should already exist. Please log an error and disconnect if it isn't. Please research if we need to use separate channels for each queue though I do not think so. Perhaps [this](https://www.rabbitmq.com/docs/channels) will be useful.
- [ ] On disconnect, we need to call [cancel](https://amqp-node.github.io/amqplib/channel_api.html#channel_cancel) for the queue of that agent. To do this, we will need `consumerTag` we get when we start consuming. Please store these in memory somehow, likely also with Cache like we are already doing for the Agent list.

apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -0,0 +1,63 @@
...
+      this.logger.error(`Error during agent socket connection: ${client.id}`, 'AgentGateway');
+      client.disconnect();
+    }
+  }
+
+  async handleDisconnect(@ConnectedSocket() client: Socket): Promise<void> {
+    const agentId = client?.agent?.sub;
+
+    if (agentId) {
+      await this.rabbitService.deleteQueue(agentId);
```
IgorKrupenja: Please see my comment above about `setupQueue`.

IgorKrupenja (changes requested):

apps/dmr-server/src/modules/gateway/dmr.server.gateway.ts
sherifbai: okay, I will delete auth guard

IgorKrupenja (changes requested):
- [ ] One of the GitHub Actions is failing. Please check why. Can we fix this now?
- [ ] Please also resolve the only small conflict with the base branch.

PR #37, opened 2025-06-13
34: Retry/reconnect logic for RabbitMQ connection

### LINKED ISSUE #34: Retry/reconnect logic for RabbitMQ connection
**AS AN** Architect
**I WANT** I want DMR server to retry connections and re-connect to RabbitMQ
**SO THAT** the system is resilient and can handle connection issues.

## Acceptance criteria

- [ ] DMR server retries connections if it cannot reach RabbitMQ.
- [ ] If connection to RabbitMQ drops, DMR server attempts to restore connection.
- [ ] Research how to better implement this. `ampqlib` does not seem to have a built-in mechanism. Should we built something ourselves? Or should we just use Rascal instead? It seems to offer a [fairly easy](https://www.npmjs.com/package/rascal/v/3.2.1#connection) way to do this.
- [ ] Please check this in a test if not too difficult.

aNazirov:
Implemented retrry/reconnect logic for rabbit client, and сovered with unit tests as much as possible.


apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.spec.ts
```diff
@@ -28,11 +31,13 @@ vi.mock('amqplib', async () => {
   };
 });
 
+import { SchedulerRegistry } from '@nestjs/schedule';
 import * as amqplib from 'amqplib';
```
IgorKrupenja: Please move imports to the top of the file. This feels a bit confusing.

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
```diff
@@ -1,43 +1,97 @@
...
+
+      this.scheduleReconnect();
+    }
+  }
+
+  async onModuleDestroy() {
+    this.schedulerRegistry.deleteInterval(this.RECONNECT_INTERVAL_NAME);
+
+    this._connection?.removeAllListeners();
+    await this._connection?.close();
```
IgorKrupenja, line 36: Thank you for thinking about this!

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
```diff
@@ -1,43 +1,97 @@
...
   }
 
+  private onClose() {
+    this.logger.warn('Connection closed. Reconnecting...');
+    this.scheduleReconnect();
+  }
+
+  private scheduleReconnect(): void {
+    if (this.schedulerRegistry.doesExist('interval', this.RECONNECT_INTERVAL_NAME)) {
+      return; // already planned
```
IgorKrupenja: What do you mean by "already planned" here?

IgorKrupenja (changes requested):

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
aNazirov: Sorry, a litle bit incorrect comment, It should be "Interval already exist", I know it's unnecessary, but for improve readability

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.spec.ts
aNazirov: Moved the Schedule Registry import, but I have to leave the `amqplib` import here - it’s important for the test.

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
IgorKrupenja: Yeah, now much better, thanks!

IgorKrupenja (approved):

PR #41, opened 2025-06-13
32: Receiving agent list from DMR Server

### LINKED ISSUE #32: Receiving agent list from DMR Server
**AS AN** Architect
**I WANT** DMR Agent to receive and store other agent data
**SO THAT** DMR Agents have the metadata and public keys they need to communicate with other DMR Agents

## Related tasks

- Depends on DMR Server task: https://github.com/buerokratt/DMR/issues/27

## Acceptance criteria

- [ ] DMR Agent has a very simple way to store other agent list in memory. Suggest using the same Cache mechanism we already have in DMR Server.
- [ ] DMR Agent is listening to full and partial list events defined in the DMR Server task ( https://github.com/buerokratt/DMR/issues/27) and is able to update its agent list when it receives the events.

furqattrade:
Related task: https://github.com/buerokratt/DMR/issues/32


apps/dmr-agent/src/modules/agents/agents.service.ts
```diff
@@ -0,0 +1,111 @@
...
+    });
+
+    socket.on(AgentEventNames.PARTIAL_AGENT_LIST, (data: IAgentList) => {
+      void this.handlePartialAgentListEvent(data);
+    });
+  }
+
+  private async handleFullAgentListEvent(data: IAgentList): Promise<void> {
+    try {
+      const activeAgents = data.response.filter((agent) => !agent.deleted);
```
IgorKrupenja: We do not even need to check for `deleted` when getting the full list, we can just overwrite the list we have in cache. In this case the list we get from the server will not have any deleted agents. Sorry if it was not clear from the Slack thread.

apps/dmr-agent/src/modules/agents/agents.service.ts
```diff
@@ -0,0 +1,111 @@
...
+      this.logger.error(`Error handling full agent list: ${errorMessage}`);
+    }
+  }
+
+  private async handlePartialAgentListEvent(data: IAgentList): Promise<void> {
+    try {
+      const currentAgents = (await this.cacheManager.get(this.AGENTS_CACHE_KEY)) || [];
+
+      const agentMap: Map<string, IAgent> = new Map();
+      if (Array.isArray(currentAgents)) {
```
IgorKrupenja: `Array.isArray` check is redundant here, you have an empty array as a fallback value above. Please remove this.

apps/dmr-agent/src/modules/agents/agents.service.ts
```diff
@@ -0,0 +1,111 @@
...
+        currentAgents.forEach((agent: IAgent) => agentMap.set(agent.id, agent));
+      }
+
+      if (data.response && Array.isArray(data.response)) {
+        data.response.forEach((agent: IAgent) => {
+          if (agent.id) {
+            if (agent.deleted) {
+              if (agentMap.has(agent.id)) {
```
IgorKrupenja: Can we use just one level of nesting for `if`s? Like:

```ts
if (agent.id && agent.deleted && agentMap.has(agent.id)) { ... }
```

Or even:

```ts
if (agentMap.has(agent.id) && agent.deleted) { ... }
```

apps/dmr-agent/src/modules/agents/agents.service.ts
```diff
@@ -0,0 +1,111 @@
...
+  async getAllAgents(): Promise<IAgent[]> {
+    try {
+      const agents: IAgent[] = (await this.cacheManager.get(this.AGENTS_CACHE_KEY)) || [];
+      return agents;
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : JSON.stringify(error);
+      this.logger.error(`Error getting agents: ${errorMessage}`);
+      return [];
+    }
+  }
```
IgorKrupenja: Please remove this method for now. I do not think we will ever need to get the full list of agents from memory in the agent code. If we decide that we want this at some point, we can easily add this functionality later.

libs/shared/src/interfaces/agent-events.interface.ts
```diff
@@ -0,0 +1,12 @@
...
+export enum AgentEventNames {
+  FULL_AGENT_LIST = 'dmr:agents:full-list',
+  PARTIAL_AGENT_LIST = 'dmr:agents:partial-list',
+}
+export interface IAgent extends IAgentConfig {
+  deleted?: boolean;
+}
+export interface IAgentList {
+  response: IAgent[];
+}
```
IgorKrupenja: Please move these into a separate file, naming it `agent.interface.ts` or similar.

libs/shared/src/interfaces/agent-events.interface.ts
IgorKrupenja, line 1: Since this file should only have a an enum in it (see my comment below), please move it to `enums/agent-event-names.enum.ts`.

IgorKrupenja (changes requested):

apps/dmr-agent/src/modules/agents/agents.service.ts
```diff
@@ -57,26 +56,22 @@ export class AgentsService implements OnModuleInit {
...
-              }
-            } else {
-              agentMap.set(agent.id, agent);
-            }
-          }
-        });
-      }
+      currentAgents.forEach((agent: IAgent) => agentMap.set(agent.id, agent));
+
+      const responseItems = Array.isArray(data.response) ? data.response : [];
```
IgorKrupenja, line 84: - [ ] Can we actually do some validation, kind of like we are [doing here](https://github.com/buerokratt/DMR/blob/dev/apps/dmr-server/src/modules/centops/centops.service.ts#L63-L71)? What do you think?
- [ ] If so, please also add this validation for full agent event.

IgorKrupenja (changes requested):

libs/shared/src/dtos/agent.dto.ts
```diff
@@ -0,0 +1,28 @@
...
+  created_at: string;
+
+  @IsString()
+  @IsOptional()
+  updated_at: string;
+
+  @IsBoolean()
+  @IsOptional()
+  deleted?: boolean;
+}
```
IgorKrupenja, line 28: This is not correct unfortunately. We would send the object with properties in `camelCase` from DMR server to agent. So exactly like in `ClientConfigDto` + optional `deleted`. Please change this.

libs/shared/src/dtos/agent.dto.ts
```diff
@@ -0,0 +1,28 @@
+import { IsBoolean, IsNotEmpty, IsOptional, IsString } from 'class-validator';
+import { IAgent } from '../interfaces/agent.interface';
+
+export class AgentDto implements IAgent {
+  @IsString()
```
IgorKrupenja: Please use `IsUUID` instead.

IgorKrupenja (changes requested):

libs/shared/src/dtos/agent.dto.ts
furqattrade: yes it was extended from centofs.interface i fixed agent dtos I think this also should be fxied 
image

aNazirov: @furqattrade  https://github.com/buerokratt/DMR/issues/13

PR #49, opened 2025-06-17
39, 44, 52: Encrypt and send messages to DMR server, Receive and decrypt messages from DMR server, Tests are failing in CI due to env variable issues

### LINKED ISSUE #39: Encrypt and send messages to DMR server
**AS AN** Architect
**I WANT** to send encrypted messages with proper metadata to DMR Server
**SO THAT** DMR server cannot see the message content but can use the metadata to process them properly

## Acceptance criteria

- [ ] DMR Agent has a service (or similar) that encrypts and sends messages to DMR Server over socket.io.
- [ ] The messages should have the following structure:

```js
{
	id: UUID/string // Message ID, for now, generate a UUID with crypto.randomUUID()
	timestamp: string // Use a current timestamp
	senderId: UUID/string // This agent ID, from AGENT_ID env variable
	recipientId: UUID/string // Recipient agent ID
	payload: string // Encrypted payload, see below
	type: MessageType // See below
}
```

- [ ] The method(s) to encrypt/send messages accept the following inputs: `senderId` and `messages`. `messages` are encrypted as `payload`. For now, assume that `messages` is just an array of strings.
- [ ] `MessageType` is an enum or type in `shared` library. Currently it only has one value, 'Message' or similar.
- [ ] Before sending the message, the `payload` needs to be encrypted using [JWE](https://en.wikipedia.org/wiki/JSON_Web_Encryption).
	- [ ] Use [jose](https://www.npmjs.com/package/jose) NPM package for this.
	- [ ] First **sign** the payload with Agent's **private key**, from `AGENT_PRIVATE_KEY` env variables
	- [ ] Then, encrypt the payload with **recipient's public key**. Get this public key from memory as implemented in https://github.com/buerokratt/DMR/issues/32. If that task is not ready, hardcode some value for now and add a `TODO` comment.
- [ ] Re-usable things like Message DTO/interface and socket.io event name are in the `shared` library.
- [ ] For now, we do not yet have the logic to actually initiate sending a message. However,  please add some meaningful tests if possible.

### LINKED ISSUE #44: Receive and decrypt messages from DMR server
**AS AN** Architect
**I WANT** to receive encrypted messages from DMR Server and decrypt them
**SO THAT** I can forward these messages to the other service(s) in the client cluster

## Notes

- Related task: https://github.com/buerokratt/DMR/issues/39

## Acceptance criteria

- [ ] DMR Agent listens to incoming messages via socket.io to the DMR Server.
- [ ] On receiving a message, it validates the format. Format description is in this task, re-use the DTO/interface created there: https://github.com/buerokratt/DMR/issues/39 .
- [ ] If the format is correct, the payload signature is checked with the sender Agent public key that is looked up from agent list form memory (by `senderId`).
- [ ] The payload is decrypted using this agent's **private key** from an env variable.
- [ ] More details on encryption in the related task: https://github.com/buerokratt/DMR/issues/39
- [ ] If any of these steps fail, an error is logged for now. We may change this logic later, there is going to be a separate task for it.
- [ ] If everything is successful, a success message is logged for now.

### LINKED ISSUE #52: Tests are failing in CI due to env variable issues
Tests started failing in CI after this PR: https://github.com/buerokratt/DMR/pull/49

This is because of the issues with env variables. For now, let's just use `.env.example` in CI.

There is going to be a follow-up task to use GitHub env variables/secrets at some later point.

aNazirov:


aNazirov:
Screenshot 2025-06-17 142339

It works in local

apps/dmr-agent/src/common/config/agent.config.ts
```diff
@@ -0,0 +1,20 @@
+import { Utils } from '@dmr/shared';
+import { ConfigType, registerAs } from '@nestjs/config';
+import Joi from 'joi';
+
+export const AGENT_CONFIG_TOKEN = Symbol('AGENT_CONFIG_TOKEN');
+
+const variables = Utils.validateObject(
+  {
+    uuid: process.env.AGENT_ID,
```
IgorKrupenja: Please rename `uuid` to just `id`. This will be consistent with other places (including code outside of DMR) and makes searching a bit easier.

apps/dmr-agent/src/modules/agents/agents.service.ts
```diff
@@ -120,4 +131,34 @@ export class AgentsService implements OnModuleInit {
       return null;
     }
   }
+
+  async encryptMessagePayloadFromExternalService(
+    message: ExternalServiceMessageDto,
+  ): Promise<AgentMessageDto | null> {
+    try {
+      const uuid = crypto.randomUUID();
+      const recipient = await this.getAgentById(message.recipientId);
```
IgorKrupenja, line 161: What if `recipient` is `null`? Don't we want to check for this?

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
```diff
@@ -49,10 +49,17 @@ export class RabbitMQService implements OnModuleInit, OnModuleDestroy {
 
     this._connection.on('close', () => this.onClose());
     this._connection.on('error', (error: Error) => {
+      this.logger.error(error);
       this.logger.error(`RabbitMQ connection error: ${error.message}`);
```
IgorKrupenja: Please log the error only once.

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
```diff
@@ -49,10 +49,17 @@ export class RabbitMQService implements OnModuleInit, OnModuleDestroy {
...
     });
 
     this._channel = await this._connection.createChannel();
+
+    this._channel.on('error', (error: Error) => {
+      this.logger.error(error);
+      this.logger.error(`RabbitMQ chanel error: ${error.message}`);
```
IgorKrupenja: Please log the error only once.

apps/dmr-agent/src/modules/agents/agents.service.ts
```diff
@@ -120,4 +152,63 @@ export class AgentsService implements OnModuleInit {
...
+      this.logger.error(`Error encrypting message: ${errorMessage}`);
+      return null;
+    }
+  }
+
+  async decryptMessagePayloadFromDMRServer(
+    message: AgentEncryptedMessageDto,
+  ): Promise<AgentDecryptedMessageDto | null> {
+    try {
+      const sender = await this.getAgentById(message.senderId);
```
IgorKrupenja, line 195: What if `recipient` is `null`? Don't we want to check for this?

libs/shared/src/utils/decrypt-payload.ts
```diff
@@ -0,0 +1,26 @@
...
+  const recipientPrivateKey = await importPKCS8(recipientPrivateKeyString, 'RSA-OAEP');
+
+  // 3. Decrypt the JWE
+  const { plaintext } = await compactDecrypt(jwe, recipientPrivateKey);
+  const signedJWT = decoder.decode(plaintext);
+
+  // Load agent's public key (for signing)
+  const senderPublicKey = await importSPKI(senderPublicKeyString, 'RS256');
+
+  // Sign the payload
```
IgorKrupenja: Please remove this comment.

libs/shared/src/dtos/agent-encrypted-message.dto.ts
```diff
@@ -0,0 +1,27 @@
...
+  @IsNotEmpty()
+  recipientId: string; // Recipient agent ID
+
+  @IsString()
+  @IsNotEmpty()
+  payload: string;
+
+  @IsEnum(MessageType)
+  type: MessageType;
+}
```
IgorKrupenja, line 8: Can we have slightly less code duplication here between DTOs?

1. One option is to have a base class and then something like:

```ts
import { IsEnum, IsNotEmpty, IsString, IsUUID } from 'class-validator';
import { MessageType } from '../enums';

export abstract class AgentBaseMessageDto {
  @IsUUID()
  @IsNotEmpty()
  id: string; // Message ID, for now, generate a UUID with crypto.randomUUID()

  @IsString()
  @IsNotEmpty()
  timestamp: string; // Use a current timestamp

  @IsUUID()
  @IsNotEmpty()
  senderId: string; // This agent ID

  @IsUUID()
  @IsNotEmpty()
  recipientId: string; // Recipient agent ID

  @IsEnum(MessageType)
  type: MessageType;
}

// some other file
import { IsNotEmpty, IsString } from 'class-validator';
import { AgentBaseMessageDto } from './agent-base-message.dto';

export class AgentEncryptedMessageDto extends AgentBaseMessageDto {
  @IsString()
  @IsNotEmpty()
  payload: string;
}

// some other file
import { IsNotEmpty, IsString } from 'class-validator';
import { AgentBaseMessageDto } from './agent-base-message.dto';

export class AgentDecryptedMessageDto extends AgentBaseMessageDto {
  @IsString({ each: true })
  @IsNotEmpty()
  payload: string[];
}
```

2. Another possible option is to install `@nestjs/swagger` and use `PickType`: https://docs.nestjs.com/openapi/mapped-types

libs/shared/src/utils/decrypt-payload.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  // Verify signature and extract payload
+  const { payload } = await jwtVerify<{ data: string[] }>(signedJWT, senderPublicKey);
+
+  return payload;
+};
```
IgorKrupenja, line 24: Can we actually add some meaningful tests at this point?

libs/shared/src/utils/encrypt-payload.ts
```diff
@@ -0,0 +1,27 @@
...
+  // Load recipient's public key (for encryption)
+  const recipientPublicKey = await importSPKI(recipientPublicKeyString, 'RSA-OAEP');
+
+  // 5. Encrypt the signed JWT
+  const jwe = await new CompactEncrypt(encoder.encode(jwt))
+    .setProtectedHeader({ alg: 'RSA-OAEP', enc: 'A256GCM' })
+    .encrypt(recipientPublicKey);
+
+  return jwe;
+};
```
IgorKrupenja, line 27: Can we actually add some meaningful tests at this point?

IgorKrupenja (changes requested):
Please see below.

apps/dmr-agent/src/modules/agents/agents.service.ts
aNazirov: Thank you for mentioning, I missed it

apps/dmr-agent/src/modules/agents/agents.service.ts
aNazirov: Thank you for mentioning, I missed it

libs/shared/src/utils/decrypt-payload.ts
aNazirov: Done

libs/shared/src/utils/encrypt-payload.ts
aNazirov: Done

libs/shared/src/dtos/agent-encrypted-message.dto.ts
aNazirov: Done

libs/shared/src/utils/decrypt-payload.ts
aNazirov: Done

apps/dmr-agent/src/modules/agents/agents.service.ts
aNazirov: Done

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
aNazirov: Fixed

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
aNazirov: Fixed

apps/dmr-agent/src/modules/agents/agents.service.ts
aNazirov: Done

apps/dmr-agent/src/common/config/agent.config.ts
aNazirov: Done

IgorKrupenja (approved):

IgorKrupenja:
Related tasks:

https://github.com/buerokratt/DMR/issues/52
https://github.com/buerokratt/DMR/issues/44
https://github.com/buerokratt/DMR/issues/39

PR #50, opened 2025-06-17
27: Distributing DMR Agent list

### LINKED ISSUE #27: Distributing DMR Agent list
**AS AN** Architect
**I WANT** DMR Server to distribute DMR agent list to DMR Agents
**SO THAT** DMR Agents have the metadata and public keys they need to communicate with other DMR Agents

## Related tasks

- Depends on: https://github.com/buerokratt/DMR/issues/15 and https://github.com/buerokratt/DMR/issues/21

## Acceptance criteria

- [ ] Right after DMR Agent establishes a Socket.io connection with the server, the DMR Server should send the **full** list of DMR Agents it has in memory (the same it gets from CentOps) to this Agent as a Socket.io event. The event has a reasonable name.
- [ ] When DMR Server receives a updated list of agents from CentOps, it sends this updated list to **every** DMR Agent currently connected to it. Since we already have logic to detect deleted agents (needed for queues), I suggest:
  - [ ] We only send a **partial** list: deleted and updated (by checking `updatedAt`) agent info. For deleted agents, send them with a new optional `deleted` boolean property in the `ClientConfigDto`.
  - [ ] If there are no updates, do not send anything. This should be the most common case in production, changes to Agent list should be fairly rare.
  - [ ] The event has a reasonable name, distinct from the full list event name.

sherifbai:
Related task: https://github.com/buerokratt/DMR/issues/27


apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -59,6 +67,14 @@ export class AgentGateway implements OnGatewayConnection, OnGatewayDisconnect {
     this.logger.log(`Agent disconnected: ${agentId} (Socket ID: ${client.id})`);
   }
 
+  @OnEvent(CentOpsEvent.UPDATED)
+  onAgentConfigUpdate(data: GetDifference): void {
+    this.server.emit(AgentEventNames.PARTIAL_AGENT_LIST, data.removed);
+    this.server.emit(AgentEventNames.PARTIAL_AGENT_LIST, data.added);
```
IgorKrupenja: Can we emit this as a single event? Something like:

```ts
this.server.emit(AgentEventNames.PARTIAL_AGENT_LIST, [...data.added, ...data.removed]);
```

apps/dmr-server/src/modules/centops/interfaces/get-difference.interface.ts
```diff
@@ -0,0 +1,6 @@
+import { AgentDto } from '@dmr/shared';
+
+export interface GetDifference {
+  added: AgentDto[];
+  removed: AgentDto[];
```
IgorKrupenja: Please rename `removed` to `deleted` to match the DTO.

apps/dmr-server/src/modules/centops/interfaces/get-difference.interface.ts
```diff
@@ -0,0 +1,6 @@
+import { AgentDto } from '@dmr/shared';
+
+export interface GetDifference {
```
IgorKrupenja: I would also give a more descriptive name here. `CentOpsConfigurationDifference` or something similar? Please also rename the file then.

IgorKrupenja (changes requested):

IgorKrupenja (approved):

PR #51, opened 2025-06-17
16: Process incoming messages

### LINKED ISSUE #16: Process incoming messages
**AS AN** Architect
**I WANT** DMR Server to process incoming messages
**SO THAT** they are added to the RabbitMQ queue and can be consumed

## Notes

- Depends on https://github.com/buerokratt/DMR/issues/15
- Acknowledgement and retry logic for messages is out of scope for now. We will create separate tasks for this later.

## Acceptance criteria

- [ ] DMR server receives the messages in the following format from DMR Agents over open socket.io connection with Agent. A shared DTO/interface is used as created in [this task](https://github.com/buerokratt/DMR/issues/39).

```js
{
	id: UUID/string
	timestamp: string
	senderId: UUID/string // This is sender agent ID
	recipientId: UUID/string // Recipient agent ID
	payload: string // Encrypted payload
	type: MessageType // Please use the type implemented in https://github.com/buerokratt/DMR/issues/15
}
```

- [ ] DMR server validates the message. We need to check that the fields are present, `id`, `type` and `timestamp` valid. Also check that `senderId` and `recipientId` are actually in the list of Agents stored in memory.
- [ ] If message is not valid, it is sent to a separate `validation-failures` (or some similar name) queue in RabbitMQ. 
- [ ] `validation-failures` queue is a single queue in Rabbit, separate from per-Agent DLQs. Please configure this queue in Rabbit as well. Since this is a single queue and not created dynamically, perhaps it is better to create it in RabitMQ config files. If so, please consider putting it into a separate `rabbitmq` folder.
- [ ] Invalid messages sent to the `validation-failures` queue have the same basic format as incoming message above but with `receivedAt`, `errors` and perhaps something else you feel necessary. `errors` is an array with errors, like `MISSING_PAYLOAD`, `INVALID_SENDER_ID` and so on.
- [ ] If message is valid, it is sent to the normal Agent queue. Please add `receivedAT` and anything else you feel necessary.
- [ ] Messages sent to both normal agent queues and the validation failure queues should be set to **persistent**: https://www.rabbitmq.com/docs/persistence-conf#cq-v1

furqattrade:
Related task: https://github.com/buerokratt/DMR/issues/16

Static queue also created when docker is fired 
image


apps/dmr-server/src/libs/rabbitmq/rabbitmq-message.service.ts
```diff
@@ -0,0 +1,143 @@
...
+  private readonly logger = new Logger(RabbitMQMessageService.name);
+  private readonly VALIDATION_FAILURES_QUEUE = 'validation-failures';
+
+  constructor(private readonly rabbitMQService: RabbitMQService) {
+    void this.setupValidationFailuresQueue();
+  }
+
+  private async setupValidationFailuresQueue(): Promise<void> {
+    try {
+      const success = await this.rabbitMQService.setupQueue(this.VALIDATION_FAILURES_QUEUE);
```
IgorKrupenja: There are a couple issues here:

- [ ] `setupQueue` also creates a DLQ with TTL. We do not really want to do this. The `validation-failures` is kind of an error queue anyway, the message from there should not move into any other queue from there.
- [ ] We want to have a configurable TTL for this queue as well, I forgot to mention this. After the TTL passes, the messages should be purged from this queue, similar to DLQs. Please create a new env variable for this.
- [ ] Are you sure it is better to create this static queue here in application logic and not via a [RabbitMQ configuration file](https://www.rabbitmq.com/docs/configure)? Can you discuss this with Bobur?

apps/dmr-server/src/libs/rabbitmq/rabbitmq-message.service.ts
```diff
@@ -0,0 +1,143 @@
...
+      const failureMessage: ValidationFailureMessageDto = {
+        // Try to preserve original message fields if possible
+        id: String(messageId),
+        timestamp: this.getPropertySafely(originalMessage, 'timestamp', new Date().toISOString()),
+        senderId: this.getPropertySafely(originalMessage, 'senderId', 'unknown'),
+        recipientId: this.getPropertySafely(originalMessage, 'recipientId', 'unknown'),
+        payload: isValidObject
+          ? this.getPropertySafely(originalMessage, 'payload', JSON.stringify(originalMessage))
+          : JSON.stringify(originalMessage),
+        type: this.getPropertySafely<string | undefined>(originalMessage, 'type', undefined),
```
IgorKrupenja: Honestly, I do not like this. If there is a validation error, I want to preserve the broken message as is. Maybe we can save the messages into the `validation-failures` queue like this:

```js
{
   id: 'some-uuid' // original message UUID OR crypto.randomUUID() if original is missing
   errors: [...],
   receivedAt: [...],
   message: { ... } // or originalMesssage
}
```

This way we can store the message even if it is just `null`, `undefined`, `number` etc.

apps/dmr-server/src/libs/rabbitmq/rabbitmq-message.service.ts
```diff
@@ -0,0 +1,143 @@
...
+
+        // Add validation failure specific fields
+        receivedAt,
+        errors,
+        originalMessageId: isValidObject ? String(messageId) : undefined,
```
IgorKrupenja: I do not think we need this if we do like I have suggested above.

apps/dmr-server/src/modules/gateway/agent.spec.ts
```diff
@@ -468,4 +420,112 @@ describe('AgentGateway', () => {
...
+      mockMessageValidatorService.validateMessage.mockResolvedValueOnce(validatedMessage);
+
+      await gateway.handleMessage(client, complexMessageData);
+
+      expect(mockMessageValidatorService.validateMessage).toHaveBeenCalledWith(complexMessageData);
+      expect(loggerSpy).toHaveBeenCalledWith(
+        `Received valid message from agent ${validatedMessage.senderId} to ${validatedMessage.recipientId} (ID: ${validatedMessage.id})`,
+      );
+    });
+  });
```
IgorKrupenja, line 377: Thank you for adding tests! However, is it possible to move them into a separate spec file for the new service, `message-validator.service.spec.ts`?

libs/shared/src/enums/validation-error-type.enum.ts
```diff
@@ -0,0 +1,10 @@
+export enum ValidationErrorType {
+  MISSING_FIELD = 'MISSING_FIELD',
```
IgorKrupenja: This is not used. Was it supposed to be used somewhere? If not, please remove this.

apps/dmr-server/src/modules/gateway/message-validator.service.ts
```diff
@@ -0,0 +1,170 @@
...
+
+        return message;
+      } catch (agentError) {
+        if (agentError instanceof BadRequestException) {
+          validationErrors.push(
+            this.rabbitMQMessageService.createValidationError(
+              ValidationErrorType.INVALID_AGENT,
+              agentError.message,
+            ),
+          );
```
IgorKrupenja: Please split this into two:

- [ ] We want separate errors for invalid sender and recipient. You can use `INVALID_SENDER_ID` and `INVALID_RECIPIENT_ID` for these. They are currently unused.
- [ ] Also remove `ValidationErrorType.INVALID_AGENT` then, it will be unused.

apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -59,8 +62,22 @@ export class AgentGateway implements OnGatewayConnection, OnGatewayDisconnect {
...
-  @SubscribeMessage('messageToDMR')
-  handleMessage(@ConnectedSocket() client: Socket, @MessageBody() data: string): void {
-    this.logger.log(`${client.id} sent message to DMR: ${data}`);
+  @SubscribeMessage(AgentEventNames.MESSAGE_TO_DMR_SERVER)
+  async handleMessage(
+    @ConnectedSocket() client: Socket,
+    @MessageBody() data: unknown,
+  ): Promise<void> {
+    try {
+      const validatedMessage = await this.messageValidator.validateMessage(data);
```
IgorKrupenja: Can we slightly change how we use `validateMessage`?

- [ ] We only **validate** the message in `validateMessage` and return either a normal message DTO or a validation failure DTO.
- [ ] If validation is fine, we call `rabbitMQMessageService.sendValidMessage` here in this method.
- [ ] If validation fails, we call `this.rabbitMQMessageService.sendValidationFailure` here in this method.

This way, the concerns are separated better, the code is more readable and we avoid having to call `this.rabbitMQMessageService.sendValidationFailure` in multiple places.

apps/dmr-server/src/libs/rabbitmq/rabbitmq-message.service.ts
IgorKrupenja, line 1: Is it possible to add meaningful tests here as well?

apps/dmr-server/src/libs/rabbitmq/rabbitmq-message.service.ts
```diff
@@ -0,0 +1,143 @@
...
+      return false;
+    }
+  }
+
+  createValidationError(type: ValidationErrorType, message: string): ValidationErrorDto {
+    return {
+      type,
+      message,
+    };
+  }
```
IgorKrupenja: Please move this to `message-validator.service.ts`

IgorKrupenja (changes requested):

apps/dmr-server/src/libs/rabbitmq/rabbitmq-message.service.ts
aNazirov: Agreed with you, we need to save the original message no matter what.

apps/dmr-server/init-rabbit.sh
```diff
@@ -0,0 +1,79 @@
...
+echo "$(date '+%Y-%m-%d %H:%M:%S') RabbitMQ setup complete, keeping container running..."
+
+
+rabbitmqctl stop
+
+
+sleep 5
+
+
+exec rabbitmq-server
```
IgorKrupenja: I had a call on this with Georg (another DevOps from Bürokratt) and unfortunately we do not like this approach. This will be difficult to configure with our Kubernetes setup.

Can we please go back to the old approach with creating a queue in the application layer in `RabbitMQMessageService`? Really sorry about this 😅 

Please keep in mind:

- [ ] Use the `RABBITMQ_VALIDATION_FAILURES_TTL` you created.
- [ ] This queue does not need a DLQ.

IgorKrupenja (changes requested):

apps/dmr-server/init-rabbit.sh
furqattrade: hi, no problem

apps/dmr-server/src/libs/rabbitmq/rabbitmq-message.service.ts
```diff
@@ -0,0 +1,160 @@
...
+        ) {
+          return String(messageWithId.id);
+        }
+      }
+      return this.generateUuid();
+    } catch {
+      this.logger.debug('Error extracting original message ID, using generated UUID instead');
+      return this.generateUuid();
+    }
+  }
```
IgorKrupenja: Can we make this even simpler? Something like:

```ts
private extractMessageId(originalMessage: unknown): string {
  try {
    // Check if originalMessage is an object with an id that's either string or number
    const id = (originalMessage as { id?: unknown })?.id;
    return id && (typeof id === 'string' || typeof id === 'number')
      ? String(id)
      : this.generateUuid();
  } catch {
    this.logger.debug('Error extracting original message ID, using generated UUID instead');
    return this.generateUuid();
  }
}
```

apps/dmr-server/src/libs/rabbitmq/rabbitmq-message.service.ts
```diff
@@ -0,0 +1,160 @@
...
+          this.VALIDATION_FAILURES_QUEUE,
+          this.rabbitMQConfig.validationFailuresTTL,
+        );
+        if (!success) {
+          this.logger.error(
+            `Failed to create validation failures queue ${this.VALIDATION_FAILURES_QUEUE}`,
+          );
+          return false;
+        }
+      }
```
IgorKrupenja: I think it is actually better to create the validation failures queue only once in `onModuleInit`. Also, I think right now `setupValidationFailuresQueue` called from `constructor` will always error out initially since the validation failures queue does not exist until we get the first validation failure.

So to summarise: I would move validation failures queue creation to `setupValidationFailuresQueue` and call it from `onModuleInit`, not from constructor.

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
```diff
@@ -127,6 +127,30 @@ export class RabbitMQService implements OnModuleInit, OnModuleDestroy {
...
+
+      return true;
+    } catch (error: unknown) {
+      if (error instanceof Error) {
+        this.logger.error(`Error while setting up queue ${queueName}: ${error.message}`);
+      }
+
+      return false;
+    }
+  }
```
IgorKrupenja, line 154: If possible, please add a simple unit test here as well.

apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -67,15 +69,72 @@ export class AgentGateway implements OnGatewayConnection, OnGatewayDisconnect {
...
+      | null
+      | undefined,
+    receivedAt: string,
+  ): Promise<void> {
+    // Ensure result and message exist before proceeding
+    if (!result || !result.message) {
+      throw new Error('Validation succeeded but no message was returned');
+    }
 
-    this.logger.log('Agent configurations updated and emitted to all connected clients');
```
IgorKrupenja, line 74: Is deleting `onAgentConfigUpdate` method intentional? 🤔 I think we need it still.

apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -67,15 +69,72 @@ export class AgentGateway implements OnGatewayConnection, OnGatewayDisconnect {
...
+        Array.isArray(errorData.validationErrors) &&
+        errorData.originalMessage !== undefined &&
+        typeof errorData.receivedAt === 'string'
+      ) {
+        await this.rabbitMQMessageService.sendValidationFailure(
+          errorData.originalMessage,
+          errorData.validationErrors,
+          errorData.receivedAt,
+        );
+      }
```
IgorKrupenja: I think this can be made even simpler, `validateMessage` will always throw us a well-formed error. Something like:

```ts
const errorData = error.getResponse() as {
  message: string;
  validationErrors: ValidationErrorDto[];
  originalMessage: unknown;
  receivedAt: string;
};

await this.rabbitMQMessageService.sendValidationFailure(
  errorData.originalMessage,
  errorData.validationErrors,
  errorData.receivedAt
);
```

IgorKrupenja (changes requested):

apps/dmr-server/src/libs/rabbitmq/rabbitmq-message.service.ts
furqattrade: Yes it was logic reated to config file somehow , yes I move them in onmoduleinit

apps/dmr-server/src/modules/gateway/agent.gateway.ts
furqattrade: I thnk it is added after my some PR and in merger time I deleted by mistake and now it is returned

IgorKrupenja (approved):

PR #56, opened 2025-06-18
53: [BUG] Changing values of TTL env variables for Rabbit can cause issues in setupQueue

### LINKED ISSUE #53: [BUG] Changing values of TTL env variables for Rabbit can cause issues in `setupQueue`
Changing values of TTL env variables for Rabbit can cause issues in `setupQueue`. This is happening because `assertQueue` will throw an error if it is passed different arguments for the previously created queue. There is `checkQueue` method but tends to throw an error and break the connection if the queue doesn’t exist.

Needs research. One option is to make an HTTP request to the RabbitMQ Management UI API.

Please also check other places where `checkQueue` is used and either remove it (I think this can be safely done if we already check for the presence of Agent in the list) or use the new approach.

aNazirov:


apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
```diff
@@ -150,10 +160,68 @@ export class RabbitMQService implements OnModuleInit, OnModuleDestroy {
   }
 
   // Do not use, may break the connection.
```
IgorKrupenja: Please remove this comment.

libs/shared/src/interfaces/rabbit-management-ui.interface.ts
```diff
@@ -0,0 +1,44 @@
...
+  default_queue_type: string;
+  protected_from_deletion: boolean;
+  tracing: boolean;
+  cluster_state: IClusterState;
+}
+
+interface IMetadata {
+  description: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  tags: any[];
```
IgorKrupenja: We can use `tags: unknown[]` to avoid the error.

libs/shared/src/interfaces/rabbit-management-ui.interface.ts
```diff
@@ -0,0 +1,44 @@
...
+  'x-message-ttl': number;
+  'x-queue-type': string;
+}
+
+export interface IVHost {
+  name: string;
+  description: string;
+  metadata: IMetadata;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  tags: any[];
```
IgorKrupenja: We can use `tags: unknown[]` to avoid the error.

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
```diff
@@ -150,10 +160,68 @@ export class RabbitMQService implements OnModuleInit, OnModuleDestroy {
...
+  async checkQueueInAllVHosts(queueName: string): Promise<boolean> {
```
IgorKrupenja: Why do we need this method? It is not used.

IgorKrupenja (changes requested):

IgorKrupenja (changes requested):
Please also merge `dev` and resolve conflicts.

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
aNazirov: Done

libs/shared/src/interfaces/rabbit-management-ui.interface.ts
aNazirov: Done

libs/shared/src/interfaces/rabbit-management-ui.interface.ts
aNazirov: Done

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
aNazirov: Removed

IgorKrupenja (approved):

IgorKrupenja:
Related task: https://github.com/buerokratt/DMR/issues/53

PR #65, opened 2025-06-18
54: Enable strict mode in TSConfig

### LINKED ISSUE #54: Enable strict mode in TSConfig
Currently, the `strict` flag is not enabled in TSConfig. This is not how we want our projects configured and can cause easily preventable issues, especially with no strict null checks.

## Acceptance criteria

- [ ] `strict` is set to `true` in `tsconfig.base.json`.
- [ ] When building the project, you will get a bunch of type errors with DTOs. Please simply use non-null assertion, this is the method recommended by NestJS:

```ts
export class ClientConfigDto {
  @IsUUID()
  @IsNotEmpty()
  id!: string; // note the !

  @IsString()
  @IsNotEmpty()
  name!: string;

  @IsString()
  @IsNotEmpty()
  authenticationCertificate!: string;

  @IsString()
  @IsNotEmpty()
  createdAt!: string;

  @IsString()
  @IsNotEmpty()
  updatedAt!: string;
}

```
	
- [ ] Please fix all other errors you get during build as well.

sherifbai:
Related task: https://github.com/buerokratt/DMR/issues/54


apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
```diff
@@ -90,6 +97,10 @@ export class RabbitMQService implements OnModuleInit, OnModuleDestroy {
   }
 
   async setupQueue(queueName: string, ttl?: number): Promise<boolean> {
+    if (!this._channel) {
+      throw new BadRequestException('Rabbit channel not defined');
+    }
```
IgorKrupenja: - [ ] I have an idea. Instead of checking `if (!this._channel)` and throwing in **several** methods here, how about we do this only **once** in the getter? Then we can do:

```ts
  async checkQueue(queueName: string): Promise<boolean> {
    try {
      await this.channel.checkQueue(queueName);

      return true;
    } catch {
      return false;
    }
  }
  
  // ... And so on in other methods
  
  get channel(): rabbit.Channel {
    if (!this._channel) {
      throw new Error('Rabbit channel not defined');
    }

    return this._channel;
  }
```

- [ ] Please also do not use `BadRequestException` here. It's more of an internal server error, I would use just `Error` or similar.

IgorKrupenja (changes requested):

IgorKrupenja (approved):

Issue #57, opened 2025-06-18
Handle duplicate connections from DMR agents


IgorKrupenja:
**AS AN** Architect
**I WANT** Drop existing connection from DMR agent on new connection attempt
**SO THAT** we correctly handle an edge case when a DMR agent attempts to open a second connection

## Notes

This **should not happen normally** but we would still want to handle this edge case.

## Acceptance criteria

- [ ] When a DMR Agent tries to establish a connection with DMR Server, the server should check if it already has an open socket.io connection with that agent. If there is such a connection, it should be dropped in favour of creating a new connection.
- [ ] Investigate how can we track connections. Maybe we can modify the cache we are already using to track consuming from Rabbit queues?
- [ ] DMR server should only drop the old connection **after** successfully authenticating the new connection attempt. Otherwise this will be prone to DOS attacks.

PR #67, opened 2025-06-18
17: Process outgoing messages

### LINKED ISSUE #17: Process outgoing messages
**Not ready, rough ideas below**

**AS AN** Architect
**I WANT** DMR Server to process outgoing messages
**SO THAT** they are sent to DMR Agents 

## Notes

- Acknowledgement and retry logic for messages is out of scope for now. We will create separate task(s) for this later.

## Acceptance criteria

- [ ] On consuming a message from a RabbitMQ queue, the DMR Server forwards the message to the DMR Agent over the open socket.io connection.
- [ ] Tests are added to cover this functionality if possible.

furqattrade:
[ticket](https://devtailor.atlassian.net/jira/software/c/projects/BD/boards/240?selectedIssue=BD-19)


libs/shared/src/enums/cent-ops-event.enum.ts
```diff
@@ -1,3 +1,4 @@
 export enum CentOpsEvent {
```
IgorKrupenja: - [ ] This enum now also has an event not related to CentOps. I would rename it to something like `DmrServerEvent` or similar.
- [ ] Please also rename the file.

apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -78,6 +78,36 @@ export class AgentGateway implements OnGatewayConnection, OnGatewayDisconnect {
...
+
+  private findSocketsByAgentId(agentId: string): Socket[] {
+    const sockets: Socket[] = [];
+    const connectedSockets = this.server.sockets.sockets;
+    connectedSockets.forEach((socket: Socket) => {
+      if (socket.agent?.sub === agentId) {
+        sockets.push(socket);
+      }
+    });
+    return sockets;
```
IgorKrupenja: I think we need to simplify this to only return one socket. We should only have **one** open socket connection per agent. We have a related task to ensure there is only one connection: https://github.com/buerokratt/DMR/issues/57

What do you think?

apps/dmr-server/src/modules/gateway/agent-message-handling.spec.ts
IgorKrupenja, line 1: I would move these tests to `agent.spec.ts`. We have so far been using one spec file per  file with code for testing.

apps/dmr-server/src/libs/rabbitmq/rabbitmq-message-forwarding.spec.ts
IgorKrupenja, line 1: I would move these tests to `rabbitmq.service.spec.ts`. We have so far been using one spec file per  file with code for testing.

IgorKrupenja (changes requested):

libs/shared/src/enums/dmr-server-event .enum.ts
IgorKrupenja, line 1: There is an extra space in the file name before `.enum`. Please fix by removing the space.

IgorKrupenja (changes requested):
Please also merge `dev`.

PR #75, opened 2025-06-20
Fix RabbitMQ tests

furqattrade:


IgorKrupenja (approved):

IgorKrupenja:
Merging with build failure which is already fixed in another PR.

PR #79, opened 2025-06-23
62: Handling outgoing messages in agent cluster

### LINKED ISSUE #62: Handling outgoing messages in agent cluster
**AS AN** Architect
**I WANT** to properly handle outgoing messages to other services in agent cluster
**SO THAT** I can forward messages from DMR server

## Acceptance criteria

- [ ] On receiving a message from DMR server, the DMR Agent should call the existing decrypt method and send the message as a POST request to a webhook endpoint.
- [ ] The endpoint will have no auth.
- [ ] Make this webhook endpoint configurable via an env variable.
- [ ] A sample value for this env variable is added to the root `docker-compose.yaml`.
- [ ] Message structure should be as described here: https://github.com/buerokratt/DMR/issues/61

furqattrade:
[TICKET](https://github.com/buerokratt/DMR/issues/62)


apps/dmr-agent/src/common/config/agent.config.ts
```diff
@@ -8,10 +8,12 @@ const variables = Utils.validateObject(
   {
     id: String(process.env.AGENT_ID),
     privateKey: String(process.env.AGENT_PRIVATE_KEY),
+    webhookEndpoint: String(process.env.AGENT_WEBHOOK_ENDPOINT || ''),
```
IgorKrupenja: Please rename this env variable to `OUTGOING_MESSAGE_ENDPOINT`. I would also rename `sendMessageToWebhook` and change error message texts in that method to also have "outgoing" in the name.

apps/dmr-agent/src/modules/agents/agents.service.ts
```diff
@@ -171,6 +174,18 @@ export class AgentsService implements OnModuleInit {
...
+    if (!this.agentConfig.webhookEndpoint) {
+      throw new Error('Webhook endpoint not configured');
+    }
+    try {
+      await firstValueFrom(this.httpService.post(this.agentConfig.webhookEndpoint, message));
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : JSON.stringify(error);
+      throw new Error(`Failed to send message to webhook: ${errorMessage}`);
+    }
+  }
```
IgorKrupenja, line 197: What about calling this method when we get a message from DMR server? Or is it implemented in another PR?

IgorKrupenja (changes requested):

apps/dmr-agent/src/modules/agents/agents.service.ts
furqattrade: I had this here https://github.com/buerokratt/DMR/pull/79/commits/3ed51ad866e78919decca64bf3a91726826f8a18
but in merging time it is modified I think, it is returned

IgorKrupenja (approved):

PR #84, opened 2025-06-25
 [BUG] Incorrect DTO is used for messages sent from server to agent #82

### LINKED ISSUE #82: [BUG] Incorrect DTO is used for messages sent from server to agent
`forwardMessageToAgent` correctly gets the message from RabbitMQ queue as `AgentMessageDto` [here](https://github.com/buerokratt/DMR/blob/e411387c93a02a3b1cca3943e0f30251c6647846/apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts#L258).

However, it is then passed on incorrectly as `AgentEncryptedMessageDto`. Please fix:

- [ ] Change the type [here in server code](https://github.com/buerokratt/DMR/blob/cd601e08bf0cc070606c4f41e20c39fee948c240/apps/dmr-server/src/modules/gateway/agent.gateway.ts#L88) to `AgentMessageDto`.
- [ ] Change the type [here in agent code](https://github.com/buerokratt/DMR/blob/eafc792808e508685bf4fdf7465aaab62bc1c8c5/apps/dmr-agent/src/modules/agents/agents.service.ts#L60-L62) to `AgentMessageDto`.
- [ ] Update [agent code here](https://github.com/buerokratt/DMR/blob/eafc792808e508685bf4fdf7465aaab62bc1c8c5/apps/dmr-agent/src/modules/agents/agents.service.ts#L129-L143) to properly handle the message in `AgentMessageDto` format.
- [ ] `AgentMessageDto` should also have optional `receivedAt` we are adding in DMR Server when receiving a message.
- [ ] Why `onRabbitMQMessage` is based off an event? Can't this be just a public method we call from `forwardMessageToAgent`? I am starting to think this would be more type-safe and we could have caught the issue earlier.

sherifbai:


apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -85,7 +79,7 @@ export class AgentGateway implements OnGatewayConnection, OnGatewayDisconnect {
   }
 
   @OnEvent(DmrServerEvent.FORWARD_MESSAGE_TO_AGENT)
-  onRabbitMQMessage(payload: { agentId: string; message: AgentEncryptedMessageDto }): void {
```
IgorKrupenja, line 88: Just curious, is there a reason we need to use an event here? Can't we just call the method in this service from `forwardMessageToAgent` directly? This would be more type-safe and maybe [less confusing](https://devtailor.slack.com/archives/C0918SMVCLQ/p1750840814726969) 😅 

Or do we have to use an event because RabbitMQ is a lib?

IgorKrupenja (changes requested):

IgorKrupenja (approved):

PR #100, opened 2025-06-26
64: Acknowledging outgoing messages

### LINKED ISSUE #64: Acknowledging outgoing messages
**AS AN** Architect
**I WANT** the outgoing messages from DMR Agent to be acknowledged
**SO THAT** RabbitMQ can drop delivered messages

## Notes

- Depends on https://github.com/buerokratt/DMR/issues/62

### DMR Agent

- [ ] On sending the message to the external endpoint implemented in https://github.com/buerokratt/DMR/issues/62:
	- [ ] If got 200 response, send some _success_ event back to DMR Server over socket.io.
	- [ ] In all other cases (some error, cannot connect to the external endpoint), send some _error_ back to DMR Server over socket.io.
- [ ] Mock the webhook endpoint with MSW for now if necessary.

### DMR Server and RabbitMQ

- [ ] On getting _success_ event from DMR Agent, send `ack` back to RabbitMQ fro that message. 
- [ ] On getting _failure_ event **or** some connection error to DMR Agent: needs research. If we do nothing, will RabbitMQ try to deliver the same message again? Or do we have [to send nack](https://amqp-node.github.io/amqplib/channel_api.html#channel_nack) explicitly?
- [ ] `{ noAck: false }` is removed from `RabbitMQService` code since we have custom ack logic.

aNazirov:


libs/shared/src/interfaces/socket-act-callback.interface.ts
```diff
@@ -0,0 +1,3 @@
+import { ISocketAckPayload } from './socket-act-payload.interface';
+
+export type ISocketActCallback = (payload: ISocketAckPayload) => void;
```
IgorKrupenja: Did you mean to use ISocketAc**k**Callback with a "k" and not "t"? If so, please rename the type and also the file.

apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -81,10 +82,14 @@ export class AgentGateway
...
+
     const emit = (event: string, ...arguments_: unknown[]) => {
       this.metricService.eventsSentTotalCounter.inc({ event, namespace: '/' });
 
-      return this.server.emit(event, ...arguments_);
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call
+      return originalEmit(event, ...arguments_);
     };
 
     this.server.emit = emit;
```
IgorKrupenja: Can't this all be something similar to this?

```ts
    this.server.emit = (event: string, ...arguments_: unknown[]) => {
      this.metricService.eventsSentTotalCounter.inc({ event, namespace: '/' });
      return Server.prototype.emit.call(this.server, event, ...arguments_) as boolean;
    };
```

IgorKrupenja (changes requested):
Extra:

- [ ] Is it reasonable to add some unit tests for this functionality now?
- [ ] Please also merge `dev`.

libs/shared/src/enums/socket-act.enum.ts
IgorKrupenja, line 1: The should be `ack` in the name, not `act`, right?

libs/shared/src/interfaces/socket-act-payload.interface.ts
IgorKrupenja, line 1: The should be `ack` in the name, not `act`, right?

IgorKrupenja (changes requested):

apps/dmr-server/src/modules/gateway/agent.gateway.ts
aNazirov: It looks better and should work. I implemented it this way and left a note in case it doesn’t behave as expected

libs/shared/src/enums/socket-act.enum.ts
aNazirov: Yep, my bad. I forgot to rename files

libs/shared/src/interfaces/socket-act-payload.interface.ts
aNazirov: Yep, my bad. I forgot to rename files

libs/shared/src/interfaces/socket-act-callback.interface.ts
aNazirov: Done

apps/dmr-server/src/modules/gateway/agent.gateway.ts
aNazirov: Renamed

libs/shared/src/enums/socket-act.enum.ts
aNazirov: Renamed

aNazirov:
> Extra:
> 
> * [x]  Is it reasonable to add some unit tests for this functionality now?
> * [x]  Please also merge `dev`.

Added unit tests

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
```diff
@@ -237,14 +237,22 @@ export class RabbitMQService implements OnModuleInit, OnModuleDestroy {
...
-              channel.ack(message);
-            } else {
-              this.logger.warn('Message is null');
+            if (!message) {
+              return this.logger.warn('Message is null');
+            }
+
+            const result = await this.forwardMessageToAgent(queueName, message);
+
+            console.log(result);
```
IgorKrupenja: Pleased remove this or use `logger.log`

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.spec.ts
```diff
@@ -417,7 +419,7 @@ describe('RabbitMQService', () => {
         capturedCallback!(mockMessage);
 
         expect(forwardSpy).toHaveBeenCalledWith(queueName, mockMessage);
-        expect(ackMock).toHaveBeenCalledWith(mockMessage);
+        // expect(ackMock).toHaveBeenCalledWith(mockMessage);
```
IgorKrupenja: I think this commented line can be removed, right? We are checking for `mockeMessage` on the line above already?

libs/shared/src/interfaces/socket-ack-callback.interface.ts
```diff
@@ -0,0 +1,3 @@
+import { ISocketAckPayload } from './socket-ack-payload.interface';
+
+export type ISocketActCallback = (payload: ISocketAckPayload) => void;
```
IgorKrupenja: Still Ac**t** in the type name.

apps/dmr-server/src/modules/gateway/agent.spec.ts
```diff
@@ -524,9 +525,9 @@ describe('AgentGateway', () => {
 
       gateway.forwardMessageToAgent('agent-123', testMessage);
 
-      expect(errorSpy).toHaveBeenCalledWith(
-        expect.stringContaining('Error forwarding RabbitMQ message to agent: Socket error'),
-      );
+      // expect(errorSpy).toHaveBeenCalledWith(
+      //   expect.stringContaining('Error forwarding RabbitMQ message to agent: Socket error'),
+      // );
```
IgorKrupenja: Why is this necessary?

apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -158,18 +159,37 @@ export class AgentGateway
...
+        message,
+      )) as ISocketAckPayload;
+
+      if (response.status === SocketAckStatus.ERROR) {
+        await this.rabbitMQMessageService.sendValidationFailure(
+          message,
+          response.errors,
+          message.receivedAt ?? new Date().toISOString(),
+        );
+      }
```
IgorKrupenja, line 184: Thanks, very good catch!

IgorKrupenja (changes requested):

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.ts
aNazirov: Done

apps/dmr-server/src/modules/gateway/agent.spec.ts
aNazirov: I don’t know, I left it because I didn’t write it, and maybe it’s necessary for something.

libs/shared/src/interfaces/socket-ack-callback.interface.ts
aNazirov: Done

apps/dmr-server/src/libs/rabbitmq/rabbitmq.service.spec.ts
aNazirov: Removed

apps/dmr-server/src/modules/gateway/agent.gateway.ts
aNazirov: Thanks

apps/dmr-server/src/modules/gateway/agent.spec.ts
IgorKrupenja: Yeah, I mean why did you commented these lines out? Is it causing issues with tests? If so, we should change the test or remove it completely — it does not make sense otherwise, since this is the only `expect` statement in the test.

IgorKrupenja (changes requested):

apps/dmr-server/src/modules/gateway/agent.spec.ts
aNazirov: I found a way to fix the error caused by this line.

aNazirov: Fixed

IgorKrupenja (approved):

PR #104, opened 2025-06-26
63: Acknowledging incoming messages

### LINKED ISSUE #63: Acknowledging incoming messages
**AS AN** Architect
**I WANT** the incoming messages to DMR Agent to be acknowledged
**SO THAT** external services in agent cluster know that the messages have reached the queue

## Notes

- Depends on https://github.com/buerokratt/DMR/issues/61
- No retry logic necessary, this is the responsibility of external services.

### DMR Agent

- [x] On receiving the REST API request for the endpoint [implemented here](https://github.com/buerokratt/DMR/issues/61), DMR Agent does not respond immediately.
- [x] Instead, it waits for the confirmation from DMR Server over socket.io. This should be possible to implement [with socket.io acknowledgements](https://socket.io/docs/v3/emitting-events/#acknowledgements). Please research if necessary.
- [x] If it cannot connect to DMR Server, DMR Agent responds with "502 Bad Gateway".
- [ ] If it cannot get a response from DMR Server within a certain timeout period, DMR Agent responds with "504 Gateway Timeout". Make this timeout configurable via an env variable, default value 2 seconds.
- [ ] A sample value for this env variable is added to the root `docker-compose.yaml`.
- [ ] If it gets a response from DMR Server, DMR Agent responds with "200 OK" or "502 Bad Gateway", depending response from DMR Server (success or failure, see below).

### DMR Server

- [ ] On receiving a message via socket.io from DMR Agent and forwarding to a RabbitMQ queue:
	- [ ] If added to the queue successfully, send some _success_ event back to DMR Agent over socket.io.
	- [ ] In all other cases (some error, cannot connect to RabbitMQ), send some _error_ back to DMR Agent over socket.io.

sherifbai:


libs/shared/src/enums/socket-act.enum.ts
```diff
@@ -0,0 +1,4 @@
+export enum SocketActEnum {
```
IgorKrupenja: The name is different from the one Bobur used in [his PR](https://github.com/buerokratt/DMR/pull/100). Please sync up with him to make sure we have only one common enum.

libs/shared/src/interfaces/socket-act-response.interface.ts
IgorKrupenja, line 1: The should be `ack` in the name, not `act`, right?

libs/shared/src/interfaces/jwt-payload.interface.ts
IgorKrupenja, line 1: Thanks for getting rid of a duplicate interface! However, can you please rename the remaining interfaces file to "jwt-payload.interface.ts" and **not** "jwt-payload.interface**s**.ts" (note the "s" for plural).

apps/dmr-agent/src/modules/messages/messages.service.ts
```diff
@@ -216,6 +227,45 @@ export class MessagesService implements OnModuleInit {
...
+        'Failed to get socket instance even though connection was reported as active',
+      );
+      throw new BadGatewayException(
+        'Failed to get socket instance even though connection was reported as active.',
+      );
+    }
+
+    try {
+      const ack = (await socket
+        .timeout(this.dmrServerConfig.ackTimeoutMs)
```
IgorKrupenja: I have actually made a mistake in the task 😄 Sorry!

- [ ] The timeout should not be for the socket.io connection to the server but for the incoming `/messages` POST request itself. Please change this, should be similar to this: https://docs.nestjs.com/interceptors#more-operators
- [ ] Please also rename the env variable. Something like `INCOMING_MESSAGE_REQUEST_TIMEOUT_MS` or similar.

IgorKrupenja (changes requested):

libs/shared/src/enums/socket-act.enum.ts
IgorKrupenja, line 1: Please also rename the file, should be `ack`, not `act`. This matches Bobur's PR.

apps/dmr-agent/src/common/interceptors/timeout.interceptor.ts
```diff
@@ -0,0 +1,28 @@
...
+@Injectable()
+export class TimeoutInterceptor implements NestInterceptor {
+  constructor(@Inject(appConfig.KEY) private readonly agentConfig: AppConfig) {}
+
+  intercept(context: ExecutionContext, next: CallHandler) {
+    return next.handle().pipe(
+      timeout(this.agentConfig.incomingMessageRequestTimeoutMs),
+      catchError((error: unknown) => {
+        if (error instanceof TimeoutError) {
+          return throwError(() => new RequestTimeoutException());
```
IgorKrupenja: Please use `GatewayTimeoutException` here. I think this [is more correct](https://www.rfc-editor.org/rfc/rfc7231#section-6.6.5), since DMR Agent is acting as a proxy here basically.

IgorKrupenja (changes requested):

IgorKrupenja (approved):

PR #111, opened 2025-06-27
108: [BUG] Namespace is not used consistently for Prometheus metrics

### LINKED ISSUE #108: [BUG] Namespace is not used consistently for Prometheus metrics
In `AgentGateway`:

1. Gateway namespace is set to `/v1/dmr-agent-events` [here](https://github.com/buerokratt/DMR/blob/39fe10c0ac854ec6abaaa2f6abb886e2063816bb/apps/dmr-server/src/modules/gateway/agent.gateway.ts#L36).
2. In `onModuleInit` we inconsistently use either `socket.nsp.name` or `'/'`. This likely means that some metrics will be recorded under `/v1/dmr-agent-events`, others under `/`.

Please:

- [ ] Fix this to use actual namespace everywhere. In the emit override we might need to use something like this, but need to double-check: 

```ts
this.metricService.eventsSentTotalCounter.inc({ event, namespace: this.server.of.name });
```

aNazirov:


IgorKrupenja (approved):

IgorKrupenja (changes requested):
Please merge dev.

IgorKrupenja (approved):

PR #112, opened 2025-06-29
70: Validate incoming messages

### LINKED ISSUE #70: Validate incoming messages
**AS AN** Architect
**I WANT** to validate incoming messages from other services in agent cluster
**SO THAT** we can detect malformed messages

## Acceptance criteria

- [ ] `ExternalServiceMessageDto` is modified to match this structure:

<details>

<summary>Click to see structure</summary>

```ts
import { MessageType } from '../enums';

export interface Message<T> {
  /**
   * Unique message UUID
   */
  id: string;

  /**
   * Recipient DMR agent UUID
   */
  recipientId: string;

  /**
   * ISO 8601 timestamp
   */
  timestamp: string;

  /**
   * Message type, for now only "ChatMessage"
   */
  type: MessageType;

  /**
   * Message payload, currently can be only ChatMessagePayload
   */
  payload: T;
}

interface ChatMessagePayload {
  chat: Chat;
  messages: ChatMessage[];
}

interface Chat {
  id: string; // UUID
  endUserFirstName?: string;
  endUserLastName?: string;
  endUserId?: string;
  endUserEmail?: string;
  endUserPhone?: string;
  customerSupportDisplayName?: string;
  created: string;
  endUserOs?: string;
  endUserUrl?: string;
}

interface ChatMessage {
  id: string; // UUID
  chatId: string; // UUID
  content?: string;
  event?: string;
  csaTitle?: string;
  authorId?: string;
  authorTimestamp: string;
  authorFirstName: string;
  authorLastName?: string;
  authorRole: string;
  forwardedByUser: string;
  forwardedFromCsa: string;
  forwardedToCsa: string;
  originalBaseId?: string;
  originalCreated?: string;
  rating?: string;
  created?: string;
  preview?: string;
  updated?: string;
  buttons?: string;
  options?: string;
}
```

</details>

- [ ] Incoming messages are validated. To validate `payload`, check what is the value of `type` and apply proper validation based on that value. Currently only `ChatMessage`.
- [ ] When creating `AgentMessageDto`, `timestamp` value needs to be set based on `timestamp` in `ExternalServiceMessageDto`.
- [ ] Make necessary changes to DMR Server code as well (e.g. `payload`).

furqattrade:
[ticket](https://github.com/buerokratt/DMR/issues/70)


apps/dmr-agent/src/modules/messages/message-validator.service.ts
```diff
@@ -0,0 +1,116 @@
...
+      }
+
+      if (error.children?.length) {
+        validationErrors.push(...this.mapValidationErrors(error.children, property));
+      }
+    }
+
+    return validationErrors;
+  }
+}
```
IgorKrupenja: Thank you for attention to detail here. However, instead of using the service, can't we just use NestJS [built-in `ValidationPipe`](https://docs.nestjs.com/techniques/validation)? It is simpler, allows us to have less code to maintain and should be good enough. I think this should be set up in a way similar to this but please double-check:

```ts
// main.ts
import { ValidationPipe } from '@nestjs/common';

async function bootstrap() {
  const app = await NestFactory.create(AppModule);
  app.useGlobalPipes(new ValidationPipe({
    whitelist: true, // strips properties not in the DTO
    forbidNonWhitelisted: true, // throws if unknown properties are present
    transform: true, // auto-transform payloads to DTO instances
  }));
  await app.listen(3000);
}
bootstrap();
```

IgorKrupenja (changes requested):

libs/shared/src/dtos/external-service-message.dto.ts
```diff
@@ -170,6 +171,7 @@ export class ChatMessagePayloadDto {
 }
 
 export class ExternalServiceMessageDto extends Message<ChatMessagePayload> {
+  @ValidateIf((obj) => obj.type === MessageType.ChatMessage)
```
IgorKrupenja, line 173: Very cool find, thanks!

apps/dmr-agent/src/modules/messages/message-validator.service.ts
furqattrade: Done and test also added

libs/shared/src/dtos/external-service-message.dto.ts
```diff
@@ -1,15 +1,178 @@
...
+  @Type(() => ChatMessageDto)
+  messages!: ChatMessageDto[];
+}
+
+export class ExternalServiceMessageDto extends Message<ChatMessagePayload> {
+  @ValidateIf((obj) => obj.type === MessageType.ChatMessage)
+  @ValidateNested()
+  @Type(() => ChatMessagePayloadDto)
+  payload!: ChatMessagePayloadDto;
 }
```
IgorKrupenja, line 177: Every class is exported from this file even though only `ExternalServiceMessageDto` is used outside of this file. Please export only `ExternalServiceMessageDto`.

libs/shared/src/dtos/external-service-message.dto.ts
```diff
@@ -1,15 +1,178 @@
...
+  @ValidateNested()
+  @Type(() => ChatDto)
+  chat!: ChatDto;
+
+  @ValidateNested({ each: true })
+  @Type(() => ChatMessageDto)
+  messages!: ChatMessageDto[];
+}
+
+export class ExternalServiceMessageDto extends Message<ChatMessagePayload> {
```
IgorKrupenja: We are using `ChatMessagePayload` interface here. Can't we just use the DTO class `ChatMessagePayloadDto` instead? If we can, please just delete the file with interfaces.

libs/shared/src/dtos/external-service-message.dto.ts
```diff
@@ -1,15 +1,178 @@
...
+  @IsString()
+  authorId?: string;
+
+  @IsString()
+  @IsNotEmpty()
+  authorTimestamp!: string;
+
+  @IsString()
+  @IsNotEmpty()
+  authorFirstName!: string;
```
IgorKrupenja: Please actually make this optional. Unfortunately, this is a mess in Burokratt 😄 I had to double-check the external DB.

libs/shared/src/dtos/external-service-message.dto.ts
```diff
@@ -1,15 +1,178 @@
...
+  @IsString()
+  authorLastName?: string;
+
+  @IsString()
+  @IsNotEmpty()
+  authorRole!: string;
+
+  @IsString()
+  @IsNotEmpty()
+  forwardedByUser!: string;
```
IgorKrupenja: Please actually make this optional. Unfortunately, this is a mess in Burokratt 😄 I had to double-check the external DB.

libs/shared/src/dtos/external-service-message.dto.ts
```diff
@@ -1,15 +1,178 @@
...
+
+  @IsString()
+  @IsNotEmpty()
+  forwardedFromCsa!: string;
```
IgorKrupenja: Please actually make this optional. Unfortunately, this is a mess in Burokratt 😄 I had to double-check the external DB.

libs/shared/src/dtos/external-service-message.dto.ts
```diff
@@ -1,15 +1,178 @@
...
+
+  @IsString()
+  @IsNotEmpty()
+  forwardedToCsa!: string;
```
IgorKrupenja: Please actually make this optional. Unfortunately, this is a mess in Burokratt 😄 I had to double-check the external DB.

apps/dmr-agent/src/modules/messages/messages.service.ts
```diff
@@ -283,19 +299,21 @@ export class MessagesService implements OnModuleInit {
         return null;
       }
 
+      const serializedPayload = JSON.stringify(message.payload);
```
IgorKrupenja: Sorry, I just don't get this. We are we serialising and deserialising? I really see no reason to do this, we should be able to encrypt anything. I would just change the type of `payload` in `encryptPayload` to `unknown` or something similar.

IgorKrupenja (changes requested):

libs/shared/src/dtos/external-service-message.dto.ts
furqattrade: Done

apps/dmr-agent/src/modules/messages/messages.service.ts
furqattrade: Yes I make type unknow an dfix this issue

libs/shared/src/dtos/external-service-message.dto.ts
IgorKrupenja: Please also change the property to be optional, i.e. `authorFirstName?: string`

libs/shared/src/dtos/external-service-message.dto.ts
IgorKrupenja: Please also change the property to be optional, i.e. `forwardedByUser?: string`

libs/shared/src/dtos/external-service-message.dto.ts
IgorKrupenja: Please also change the property to be optional, i.e. `forwardedFromCsa?: string`

libs/shared/src/dtos/external-service-message.dto.ts
IgorKrupenja: Please also change the property to be optional, i.e. `forwardedToCsa?: string`

IgorKrupenja (changes requested):

libs/shared/src/dtos/external-service-message.dto.ts
furqattrade: Done

apps/dmr-agent/src/modules/messages/messages.service.ts
```diff
@@ -169,7 +167,9 @@ export class MessagesService implements OnModuleInit {
       const outgoingMessage: ExternalServiceMessageDto = {
         id: message.id,
         recipientId: message.recipientId,
-        payload: decryptedMessage.payload,
+        timestamp: message.timestamp,
+        type: message.type,
+        payload: decryptedMessage.payload as any,
```
IgorKrupenja: Please use `as unknown` here as well.

IgorKrupenja (changes requested):

IgorKrupenja (approved):

PR #118, opened 2025-06-30
116: [BUG] Outgoing message acknowledgement logic is slightly wrong

### LINKED ISSUE #116: [BUG] Outgoing message acknowledgement logic is slightly wrong
I have missed this previously and this was not described well in the related task.

## Current

At the moment acknowledging outgoing messages always works this way:

1. In case of **any** failure on DMR agent side, the agent sends an acknowledgement back to the server with a `DECRYPTION_FAILED` error. [Agent code link](https://github.com/buerokratt/DMR/blob/929a441165c15587ce17c3120b6156d5e9048553/apps/dmr-agent/src/modules/messages/messages.service.ts#L152-L195).
2. The DMR server **always** then sends a this message to the `validation-failures` queue and a nack to RabbitMQ. [Server code link](https://github.com/buerokratt/DMR/blob/e051867e07fc254a314024965944f42a95bb90a3/apps/dmr-server/src/modules/gateway/agent.gateway.ts#L179-L190).

## Needed

We want to distinguish between two cases.

### Case 1: Decryption failed

1. If message decryption fails, DMR agent sends an acknowledgement back to the server with a `DECRYPTION_FAILED` error. As it does now, no changes.
2. The DMR server should check if the error type is `DECRYPTION_FAILED`. In this case, it should send the message to the `validation-failures` queue. It should also send a `nack` to RabbitMQ with `requeue` argument being `false`, i.e. `channel.nack(message, false, false)`.

### Case 2: Delivery failed

Any issues related to connection failures etc.

1. If DMR agent fails to deliver the message to an external service (connection down or similar), it should send an acknowledgement back to the server with some other error type, i.e. **not** `DECRYPTION_FAILED`. E.g. let's say `DELIVERY_FAILED`.
2. DMR server should check if the error type is `DELIVERY_FAILED`. In this case, it should **not** send the message to the `validation-failures` queue. It should send a `nack` to RabbitMQ with `requeue` argument being `true`, i.e. `channel.nack(message, false, true)`. RabbitMQ will then use [the default logic](https://www.rabbitmq.com/docs/quorum-queues#poison-message-handling) to attempt re-delivery, this is fine.

Additionally:

- [ ] If DMR server cannot connect to the agent or there are some connection issues, it should do as in point 2 above (nack with requeue true). The timeout for connection should be set based on an env variable. Right now it is called `INCOMING_MESSAGE_REQUEST_TIMEOUT_MS` but please rename it to something more generic as we are using it for outgoing as well. Some `MESSAGE_DELIVERY_TIMEOUT_MS` or something?

aNazirov:


IgorKrupenja (changes requested):
- [ ] Can we add some tests for this?
- [ ] Please merge dev.
- [ ] You are absolutely right in this [comment here](https://github.com/buerokratt/DMR/issues/116#issuecomment-3019162461). This env variable exists in DMR Agent only. A question to you then: what will happen if DMR agent is down and `forwardMessageToAgent` fails? Will it just error out or will we still do `channel.nack(message, false, true)`? The idea is to handle this case gracefully too.

aNazirov:
> * [ ]  Can we add some tests for this?
> * [ ]  Please merge dev.
> * [ ]  You are absolutely right in this [comment here](https://github.com/buerokratt/DMR/issues/116#issuecomment-3019162461). This env variable exists in DMR Agent only. A question to you then: what will happen if DMR agent is down and `forwardMessageToAgent` fails? Will it just error out or will we still do `channel.nack(message, false, true)`? The idea is to handle this case gracefully too.

1) Added tests
2) Done
3) It should cause an error, which is caught in the try...catch block and handled by calling channel.nack(message, false, true).

IgorKrupenja (approved):

IgorKrupenja:
@aNazirov 

3. Very good, thanks! I think this is fine then.

PR #119, opened 2025-06-30
110: [BUG] DMR Agent does not establish socket.io connection to DMR server on start

### LINKED ISSUE #110: [BUG] DMR Agent does not establish socket.io connection to DMR server on start
1. Check out this PR branch, it has docker compose fixed: https://github.com/buerokratt/DMR/pull/105
2. Run `docker compose up`.

There is nothing in logs for DMR agents, meaning they do not connect to DMR server over socket.io on start. Also, if you try sending a request to `/v1/messages` you will get 500 since agent list is empty. The agent does not get it from server since there is no connection.

furqattrade:
[ticket](https://github.com/buerokratt/DMR/issues/110)


apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -90,8 +90,6 @@ export class AgentGateway
 
       return Server.prototype.emit.call(this.server, event, ...arguments_) as boolean;
     };
-
-    this.server.emit = emit;
```
IgorKrupenja: Is this intentional? The `emit` constant is no longer used. Also, `this.server.emit = this.originalEmit` below in `onModuleDestroy` does not seem to make sense anymore. Please consult with Bobur on this if necessary.

I also noticed that lint and format GitHub Actions are not running properly. But please do not fix it in this PR, I will create a separate issue for it.

IgorKrupenja (changes requested):

PR #124, opened 2025-07-01
117: Dropping connections to agents with revoked/rotated certificates

### LINKED ISSUE #117: Dropping connections to agents with revoked/rotated certificates
**AS AN** Architect
**I WANT** DMR Server to drop connections to agents with revoked/rotated certificates
**SO THAT** we can proactively respond in case of a security breach

## Acceptance criteria

- [ ] When DMR server gets he agent list from CentOps, it should check the currently opensocket.io connections with DMR agents. It should immediately drop the connection if:
  - [ ] The connected agent is no longer in the agent list.
  - [ ] **OR** the connected agent is in the list but the public key (`authentication_certificate`) is different from the one that the agent used to establish the connection.

This is needed in a very rare scenario where there has been a security breach and some certificates are revoked or similar.

furqattrade:
[ticket](https://github.com/buerokratt/DMR/issues/117)


libs/shared/src/interfaces/jwt-payload.interface.ts
```diff
@@ -3,4 +3,5 @@ export interface JwtPayload {
   iat: number;
   exp: number;
   cat: number;
+  authenticationCertificate?: string;
```
IgorKrupenja: I quite like this approach in general, thanks!

However, I do not like how we are adding a new property to the `JwtPayload` interface. It has nothing to do with JWT payload, we do not set `authenticationCertificate` on the agent where this interface is also used. This can create confusion.

I would suggest the following instead:

- [ ] Remove `authenticationCertificate` from `JwtPayload`.
- [ ] Modify [this declaration](https://github.com/buerokratt/DMR/blob/e2ea69840a772b8225fd90d4cf50bc658f4915b6/apps/dmr-server/src/modules/gateway/types/types.d.ts#L3-L7) to something like:

```
declare module 'socket.io' {
  interface Socket {
    jwtPayload: JwtPayload;
	authenticationCertificate: string;
  }
}
```

- [ ] In `handleConnection` [here](https://github.com/buerokratt/DMR/blob/f00112a2b655c85e47ef67b2fd48c823baa6abe9/apps/dmr-server/src/modules/gateway/agent.gateway.ts#L111), do something like:

```ts
Object.assign(client, { jwtPayload, authenticationCertificate });
```

- [ ] Make all the necessary changes elsewhere, including `verifyToken`. Maybe it would make sense to extract `{ jwtPayload: JwtPayload; authenticationCertificate: string; }` into a new interface.

apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -159,10 +159,71 @@ export class AgentGateway
...
+      } else {
+        const currentAgentConfig = currentAgentMap.get(agentId);
+        if (!currentAgentConfig) {
+          shouldDisconnect = true;
+          reason = 'Agent not found in current authorized list';
+        } else if (currentAgentConfig.authenticationCertificate !== connectionCertificate) {
+          shouldDisconnect = true;
+          reason = 'Agent certificate mismatch with current configuration';
+        }
+      }
```
IgorKrupenja: I am not sure if I get this correctly. Is this `else` clause necessary here? Isn't this handled above already with `if (deletedAgentIds.has(agentId))` and `else if (certificateChangedAgentIds.has(agentId))`?

Or am I missing something?

IgorKrupenja (changes requested):

apps/dmr-server/src/modules/gateway/agent.gateway.ts
furqattrade: else block could be useful as a safety net for edge cases:
Race conditions=> Agent connects just as configuration is being updated
Cache inconsistencies => Mismatch between cached data and actual current data

libs/shared/src/interfaces/jwt-payload.interface.ts
furqattrade: Done

apps/dmr-server/src/modules/gateway/agent.gateway.ts
IgorKrupenja: Thanks, makes sense! Could you please add a short comment in code mentioning the reasons why this else block is needed then?

IgorKrupenja (changes requested):
Please see my comment above.

PR #129, opened 2025-07-01
43, 127: DMR Agent Prometheus metrics

### LINKED ISSUE #43: DMR Agent Prometheus metrics
**AS AN** Architect
**I WANT** to explore possible options for useful DMR Agent Prometheus metrics
**SO THAT** we can implement this to have better observability.

## ❗ Important

- Do not start with this before all other tasks for DMR Agent are implemented.

### Research

We want to have some [Prometheus](https://prometheus.io/) metrics for DMR Agent.

- [ ] Please research this and create a short summary in a comment for this task. The summary should answer these questions:
	- [ ] 1. What should be the list of metrics we track?
	- [ ] 2. Based on the suggested metrics from the previous point, please also suggest some useful [alerting rules](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/).

### Implementation

- [ ] Implement based on research.
- [ ] Add info on metrics and suggested alerts to the Readme.
- [ ] Add a couple simple tests to make sure that Prometheus metrics/endpoint works.

### LINKED ISSUE #127: [BUG] DMR server: errors on startup from AgentGateway.onModuleInit
1. Run `docker compose up`.

You will get this in server logs:

```log
dmr-server    | [Nest] 1  - 07/01/2025, 12:17:40 PM   ERROR [Event] Cannot read properties of undefined (reading 'name') +9994ms
dmr-server    | TypeError: Cannot read properties of undefined (reading 'name')
dmr-server    |     at Namespace.emit (/app/dist/apps/dmr-server/main.js:1800:94)
dmr-server    |     at AgentGateway.onAgentConfigUpdate (/app/dist/apps/dmr-server/main.js:1853:21)
dmr-server    |     at EventSubscribersLoader.wrapFunctionInTryCatchBlocks (/app/node_modules/.pnpm/@nestjs+event-emitter@3.0.1_@nestjs+common@11.1.3_class-transformer@0.5.1_class-validat_3b401e90ee47945f27b838d6690e18ff/node_modules/@nestjs/event-emitter/dist/event-subscribers.loader.js:97:46)
dmr-server    |     at EventEmitter.<anonymous> (/app/node_modules/.pnpm/@nestjs+event-emitter@3.0.1_@nestjs+common@11.1.3_class-transformer@0.5.1_class-validat_3b401e90ee47945f27b838d6690e18ff/node_modules/@nestjs/event-emitter/dist/event-subscribers.loader.js:73:57)
dmr-server    |     at EventEmitter.emit (/app/node_modules/.pnpm/eventemitter2@6.4.9/node_modules/eventemitter2/lib/eventemitter2.js:1011:19)
dmr-server    |     at CentOpsService.syncConfiguration (/app/dist/apps/dmr-server/main.js:2156:31)
dmr-server    |     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
dmr-server    |     at async CronJob.onTick (/app/dist/apps/dmr-server/main.js:2105:13)
Gracefully stopping... (press Ctrl+C again to force)
```

I tried hardcoding `'/v1/dmr-agent-events'` instead of `this.server.of.name` but then I get another error 😛 

```log
dmr-server    | [Nest] 1  - 07/01/2025, 12:20:10 PM   ERROR [Event] Cannot read properties of undefined (reading 'apply') +9997ms
dmr-server    | TypeError: Cannot read properties of undefined (reading 'apply')
dmr-server    |     at Server.<computed> (/app/node_modules/.pnpm/socket.io@4.8.1/node_modules/socket.io/dist/index.js:798:33)
dmr-server    |     at Namespace.emit (/app/dist/apps/dmr-server/main.js:1801:54)
dmr-server    |     at AgentGateway.onAgentConfigUpdate (/app/dist/apps/dmr-server/main.js:1853:21)
dmr-server    |     at EventSubscribersLoader.wrapFunctionInTryCatchBlocks (/app/node_modules/.pnpm/@nestjs+event-emitter@3.0.1_@nestjs+common@11.1.3_class-transformer@0.5.1_class-validat_3b401e90ee47945f27b838d6690e18ff/node_modules/@nestjs/event-emitter/dist/event-subscribers.loader.js:97:46)
dmr-server    |     at EventEmitter.<anonymous> (/app/node_modules/.pnpm/@nestjs+event-emitter@3.0.1_@nestjs+common@11.1.3_class-transformer@0.5.1_class-validat_3b401e90ee47945f27b838d6690e18ff/node_modules/@nestjs/event-emitter/dist/event-subscribers.loader.js:73:57)
dmr-server    |     at EventEmitter.emit (/app/node_modules/.pnpm/eventemitter2@6.4.9/node_modules/eventemitter2/lib/eventemitter2.js:1011:19)
dmr-server    |     at CentOpsService.syncConfiguration (/app/dist/apps/dmr-server/main.js:2156:31)
dmr-server    |     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
dmr-server    |     at async CronJob.onTick (/app/dist/apps/dmr-server/main.js:2105:13)
```

I can temporarily get rid of the errors if I comment out all code in `onModuleInit`.

aNazirov:
https://github.com/buerokratt/DMR/issues/127
https://github.com/buerokratt/DMR/issues/43


aNazirov:
@IgorKrupenja  Can we move implementing the tests into a separate task? This PR already includes quite a lot of changes, so it would be better to extract the metric tests for both dmr agent and dmr server into a dedicated shared task, I’ll complete it there.

IgorKrupenja:
> @IgorKrupenja Can we move implementing the tests into a separate task? This PR already includes quite a lot of changes, so it would be better to extract the metric tests for both dmr agent and dmr server into a dedicated shared task, I’ll complete it there.

Yes, this is a very good idea. I will create a task for this soon.

README.md
```diff
@@ -140,7 +140,85 @@ groups:
...
+
+- **`dmr_http_requests_total`** | `counter` | `method, route, status`
+  Total HTTP requests handled
+- **`dmr_http_request_duration_seconds`** | `histogram` | `method, route, status`
+  HTTP request processing time
+
+- **`dmr_http_errors_total` | `counter`** | `method, route`
+  Count of error responses (4xx/5xx)
+
+- **`dmr_agent_socket_connection_status `** | `gauge`
```
IgorKrupenja: This is a nit pick but I would rename this to `dmr_agent_socket_connection_active` (without renaming the alerts). This feels confusing to me, `status` implies a boolean, not a number.

apps/dmr-agent/.env.example
```diff
@@ -5,7 +5,8 @@ ENVIRONMENT=development
 LOGGER_COLORS=true
 LOGGER_LOG_LEVELS=error,warn,log
 
-DMR_SERVER_WEBSOCKET_URL='http://localhost:5000/v1/dmr-agent-events'
+DMR_SERVER_WEBSOCKET_URL='http://localhost:5000'
+DMR_SERVER_WEBSOCKET_NAMESPACE='/v1/dmr-agent-events'
```
IgorKrupenja: I do not think I get this. Why split this into two variables? Please also see my comment on server env variables.

Having this as one variable would also be more in line with env variables in other Bürokratt modules — and thus more familiar to DevOps.

apps/dmr-server/.env.example
```diff
@@ -7,6 +7,7 @@ LOGGER_COLORS=true
 LOGGER_LOG_LEVELS=error,warn,log
 
 WEB_SOCKET_MAX_DISCONNECTION_DURATION=120000
+WEB_SOCKET_NAMESPACE=/v1/dmr-agent-events
```
IgorKrupenja: What is the reason behinds making this configurable via an env variable instead of just hardcoding? I do not think this adds much value and cannot see a reason why this can differ from one environment to another. 

There are quite a few env variables already and I would only have the absolute necessary ones. Otherwise, it is just another thing to consider for DevOps — and needlessly increases their cognitive load.

apps/dmr-server/src/modules/gateway/agent.gateway.ts
```diff
@@ -74,32 +73,35 @@ export class AgentGateway
...
+
+      for (const socket of [...sockets.values()]) {
+        this.metricService.eventsSentTotalCounter.inc({ event, namespace: socket.nsp.name });
+      }
+
+      return originalServerEmit(event, arguments_);
     };
 
-    this.server.emit = emit;
+    this.server.emit = serverEmit;
```
IgorKrupenja, line 100: Thank you so much for fixing this! Great job 🚀

IgorKrupenja (changes requested):

apps/dmr-agent/.env.example
aNazirov: The reason for doing this was that on the client side in DMR Agent, we don’t have access to the namespace from the socket.
I also wanted to move the namespace to an environment variable on the DMR Server side to allow changing it dynamically.
But if that seems problematic, I’ve already thought of another way to implement it. Should we leave it as is or remove it from the env?

apps/dmr-server/.env.example
aNazirov: The reason for doing this was that on the client side in DMR Agent, we don’t have access to the namespace from the socket.
I also wanted to move the namespace to an environment variable on the DMR Server side to allow changing it dynamically.
But if that seems problematic, I’ve already thought of another way to implement it. Should we leave it as is or remove it from the env?

README.md
aNazirov: Renamed

apps/dmr-agent/.env.example
IgorKrupenja: > The reason for doing this was that on the client side in DMR Agent, we don’t have access to the namespace from the socket.

Aha! Yeah, I missed this. We do actually need namespace separately from the URL for metrics. 

But this still feels a bit hard to get for me. I have thought about this a bit more and checked what we do in other Burokratt modules. My suggestion is we do this:

1. Hardcode `/v1/dmr-agent-events` in **both** server and agent, just as a constant or similar.
2. Leave only one env variable `DMR_SERVER_WEBSOCKET_URL` (agent) with the value `http://localhost:5000/`, without the namespace postfix.

This way:

- There are less env variables to deal with.
- This feels less error-prone, no room for a mistake in the "/v1/dmr-agent-events" path since this is hardcoded.

Does this make sense?

aNazirov: I can just remove namespace env variable from both side, leave name space in dmr server as hardcode, and parse namespce from url in dmr agent side, it seems better. What do you think?

aNazirov: Done

apps/dmr-server/.env.example
aNazirov: Done

IgorKrupenja (approved):

PR #136, opened 2025-07-03
135: [BUG] DMR Agent is not getting agent list from server

### LINKED ISSUE #135: [BUG] DMR Agent is not getting agent list from server
Steps to reproduce:

1. Run `docker compose up`.
2. Run this to send a test message to the agent:

```sh
curl -X POST http://localhost:8074/v1/messages \
  -H "Content-Type: application/json" \
  -d '{
    "id": "b1a7e8c2-1234-4f56-9abc-1234567890ab",
    "recipientId": "d3b07384-d9a0-4c3f-a4e2-123456789abc",
    "timestamp": "2024-01-15T10:30:00.000Z",
    "type": "ChatMessage",
    "payload": {
      "chat": {
        "id": "f47ac10b-58cc-4372-a567-0e02b2c3d479",
        "created": "2024-01-15T10:00:00.000Z"
      },
      "messages": [
        {
          "id": "550e8400-e29b-41d4-a716-446655440000",
          "chatId": "f47ac10b-58cc-4372-a567-0e02b2c3d479",
          "content": "example string 2",
          "authorTimestamp": "2024-01-15T10:30:00.000Z",
          "authorRole": "end-user"
        }
      ]
    }
  }'
```

Sending the message will fail since the agent list is empty. The agent is not getting the agent list properly for the server.

There are several problems:

- `MessagesService.setupSocketEventListeners` is run in that service in `onModuleInit` once. This means that if socket connection with DMR server is not yet established at that point, the the event listeners will not be set up at all. Even when the connection eventually gets established on retry.
- There is a type mismatch for `AgentEventNames.FULL_AGENT_LIST` event between server and agent code.

IgorKrupenja:
Task: https://github.com/buerokratt/DMR/issues/135


aNazirov:
I think moving the call to setupEventListeners into WebsocketService instead of keeping it in MessagesService (as it was before) is a bit incorrect and introduces unnecessary circular dependencies into the project. What do you think?

aNazirov:
Everything else is great, thanks for your work)

apps/dmr-agent/src/modules/messages/messages.service.ts
```diff
@@ -21,42 +21,32 @@ import { Cache, CACHE_MANAGER } from '@nestjs/cache-manager';
...
-  }
-
-  onModuleDestroy() {
-    const socket = this.websocketService.getSocket();
-
-    socket.removeAllListeners();
-  }
-
-  private setupSocketEventListeners(): void {
+  setupSocketEventListeners(): void {
```
aNazirov: I think moving the call to `setupEventListeners `into `WebsocketService `instead of keeping it in `MessagesService `(as it was before) is a bit incorrect and introduces unnecessary circular dependencies into the project. What do you think?

aNazirov (changes requested):
Everything else is great, thanks for your work)

apps/dmr-agent/src/modules/messages/messages.service.ts
```diff
@@ -50,45 +49,34 @@ export class MessagesService implements OnModuleInit, OnModuleDestroy {
...
-  private setupSocketEventListeners(): void {
-    if (!this.websocketService.isConnected()) {
-      this.logger.warn(
-        'WebSocket is not connected. Will retry setting up agent event listeners when connected.',
-      );
-      return;
-    }
-
+  setupSocketEventListeners(): void {
     const socket = this.websocketService.getSocket();
```
aNazirov, line 53: Wouldn’t it be better to use the `isConnected` method from the WebsocketService?

apps/dmr-agent/src/modules/websocket/websocket.service.spec.ts
```diff
@@ -49,6 +50,10 @@ const mockMetricService = {
   messageProcessingDurationSecondsHistogram: mockHistogram,
 };
 
+const mockMessagesService = {
```
aNazirov: Is it still necessary?

apps/dmr-agent/src/modules/websocket/websocket.service.spec.ts
```diff
@@ -79,6 +84,7 @@ describe('WebsocketService', () => {
           useValue: { sign: vi.fn() },
         },
         { provide: MetricService, useValue: mockMetricService },
+        { provide: MessagesService, useValue: mockMessagesService },
```
aNazirov: Is it still necessary?

aNazirov (changes requested):

apps/dmr-agent/src/modules/messages/messages.service.ts
IgorKrupenja: Using `isConnected` was the cause of the issue here. I removed this logic above. Relying on `isConnected` could result in the listeners not being set up if socket connection is not yet established when we run this method inside `onModuleInit` here.

aNazirov: Got you

aNazirov: I can’t resolve this conversation, so that was my last comment

aNazirov (approved):

PR #159, opened 2025-07-08
154: [BUG] PARTIAL_AGENT_LIST is emitted if there are no changes

### LINKED ISSUE #154: [BUG] PARTIAL_AGENT_LIST is emitted if there are no changes
At the moment, `AgentEventNames.PARTIAL_AGENT_LIST` is needlessly emitted **every time** even if there are no changes from CentOps.

- [ ] Please fix this by emitting this event only when there are actual changes.
- [ ] Please add a test for this if possible.

sherifbai:


IgorKrupenja (changes requested):
Thanks! There are a couple more related things I forgot to mention before:

- [ ] On [this line](https://github.com/buerokratt/DMR/blob/0cce40a210c04f822142c692830bf20de0ed93fd/apps/dmr-server/src/modules/centops/centops.service.ts#L173), can you please reduce the priority of the log message from `log` to `debug`? This way we can prevent logs growing huge in size. This will also make debugging easier by not spamming this message constantly. We had problems with this before in other Burokratt modules.
- [ ] I think `DmrServerEvent.FORWARD_MESSAGE_TO_AGENT` is no longer used. Please remove it.

IgorKrupenja (approved):

PR #174, opened 2025-07-09
173: [BUG] Dependabot PRs report failures in CI

### LINKED ISSUE #173: [BUG] Dependabot PRs issues in CI
- [ ] Some formatting errors, e.g. here: https://github.com/buerokratt/DMR/actions/runs/16143993707/job/45557940467?pr=172
- [x] In some PRs **all** checks are cancelled, e.g. here: https://github.com/buerokratt/DMR/pull/171 I think this is happening because concurrency is misconfigured. In YAML config [here](https://github.com/buerokratt/DMR/blob/eafc792808e508685bf4fdf7465aaab62bc1c8c5/.github/workflows/run-checks.yml#L8-L10):

```yaml
concurrency:
  group: ${{ github.workflow }}
  cancel-in-progress: true
```

`group` is simply `github.workflow`. This means that if there are e.g. two separate PRs where the same workflow is run at the same time, the run in the first PR is going to be cancelled. Try this instead, this should only cancel running checks in the same PR (e.g. when a new commit is pushed):

```yaml
concurrency:
  group: ${{ github.ref }}-checks # -checks needed here to differentiate from e2e which is a separate workflow
  cancel-in-progress: true
```

- [x] Also, I think this PR should also belong to `nx` group: https://github.com/buerokratt/DMR/pull/170

aNazirov:
https://github.com/buerokratt/DMR/issues/173


aNazirov:
@IgorKrupenja Hi, here’s an update on the task:
Part of the issue is related to CI errors, and I believe it’s caused by `dependabot` PRs not being formatted according to our lint rules.
So, I added a new task in the pipeline that runs the `format` command and pushes the changes into the current PR.

IgorKrupenja:
@aNazirov Yeah, this is a good approach, thanks!

.github/workflows/auto-format-dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: pnpm/action-setup@v2
+        with:
+          version: 8
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: 'pnpm'
+
+      - run: pnpm install --frozen-lockfile
```
IgorKrupenja: Can you please use the reusable `./.github/actions/setup` action here. Similarly to [this](https://github.com/buerokratt/DMR/blob/eafc792808e508685bf4fdf7465aaab62bc1c8c5/.github/workflows/run-checks.yml#L56).

IgorKrupenja (changes requested):
- [ ] Can you please also have a look at the final point in the task? The one about `nx` not being part of the `nx` group. I think we need to add one more pattern [here](https://github.com/buerokratt/DMR/blob/51eb9c4f6d153aacb8afb4c95564f55fac719c61/.github/dependabot.yml#L17-L20).
- [ ] I forgot one more thing. The `dependabot.yml` was only merged by Varmo to `main` branch, not `dev`. Please also add this file in this PR, you can just copy over the file contents from the `main` branch.

.github/workflows/auto-format-dependabot.yml
aNazirov: Done

.github/workflows/auto-format-dependabot.yml
```diff
@@ -0,0 +1,34 @@
...
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: ./.github/actions/setup
+
+      - run: pnpm install --frozen-lockfile
```
IgorKrupenja: I do not think this is necessary, `.github/actions/setup` already does install.

IgorKrupenja (changes requested):

.github/workflows/auto-format-dependabot.yml
aNazirov: I've just saw, thanks for mentioning

aNazirov: Fixed

IgorKrupenja (approved):

PR #156, opened 2025-07-07
133, 140: Dockerize and implement CI for e2e tests

furqattrade:

PR #181, opened 2025-07-09
175: e2e test with two DMR servers

### LINKED ISSUE #175: e2e test with two DMR servers
# Testing DMR Server High Availability Setup

## Overview
The DMR (Distributed Message Router) system supports a high-availability configuration with two server instances. This document outlines how to test this setup.

## Test Environment
Current configuration in docker-compose.e2e.yml:
- dmr-server-1 (Primary) - Port 8075
- dmr-server-2 (Secondary) - Port 8076
- dmr-agent-a - Port 8077
- dmr-agent-b - Port 8078
- external-service-a - Port 8073
- external-service-b - Port 8074
- rabbitmq-e2e - Port 8072

### 1. Basic High Availability Test

- [ ] Verify both servers are running and healthy
- [ ] Confirm agents can connect to either server
- [ ] Test message routing works through two servers in this flow: External A -> DMR Agent A -> DMR Server A -> RabbitMQ -> DMR Server B -> DMR Agent B -> External B

### 2. Test Cases
```typescript
describe('DMR High Availability Tests', () => {
  describe('Basic Connectivity', () => {
    it('should connect agents to both servers')
    it('should route messages through primary server')
    it('should route messages through secondary server')
  })
```

aNazirov:
https://github.com/buerokratt/DMR/issues/175


package.json
```diff
@@ -64,7 +64,8 @@
     "reflect-metadata": "^0.1.13",
     "rxjs": "^7.8.0",
     "socket.io": "^4.8.1",
-    "socket.io-client": "^4.8.1"
+    "socket.io-client": "^4.8.1",
+    "uuid": "^11.1.0"
```
IgorKrupenja: Why do we need this? Is is used somewhere? Even if we need an UUID somewhere, there is an easier zero-dependency way: `crypto.randomUUID()`.

IgorKrupenja (changes requested):

package.json
aNazirov: Honestly, I’m not sure I think Furqat used it in the `advanced-scenarios.e2e.spec` file.
I agree, it would be better if I remove it and use `crypto` instead.

aNazirov: Done

IgorKrupenja: Thanks a lot for this! I actually [asked](https://github.com/buerokratt/DMR/pull/156#issuecomment-3048037459) Furqat to fix his in that PR but then forgot on re-review 😅

IgorKrupenja (approved):

PR #219, opened 2025-08-08
14: CentOps connection and auth

### LINKED ISSUE #14: CentOps connection and auth
## To do

Everything is in **DMR server** code.

- [ ] Add an env variable called `CENTOPS_API_KEY`, value some random string like `key`. This variable is required.
- [ ] Add an env variable called `CENTOPS_API_SECRET`, value some random string like `secret`. This variable is required.
- [ ] Change the sample value of `CENTOPS_CONFIGURATION_URL` to `http://localhost:3000/centops/integration/clients?pageSize=100` everywhere: docker compose, env, etc.
- [ ] When making the GET request [here](https://github.com/buerokratt/DMR/blob/ea0502a6cb5458a7d94a6a96fe683be9926c0ba3/apps/dmr-server/src/modules/centops/centops.service.ts#L105-L107), we would need to:
	- [ ] Add an `Authorization` header.
	- [ ] The value of the header should be `CENTOPS_API_KEY:CENTOPS_API_SECRET` (two env variables separated by semicolon) and **base64-encoded**.
- [ ] We want to keep [the mock](https://github.com/buerokratt/DMR/blob/aa6e304537c70e5d42244fc13ef185399f89cd6a/apps/dmr-server/src/mocks/handlers/centops.response.ts#L5) too. Please only make sure it works without issues with the newly-added `Authorization` header.

aNazirov:
https://github.com/buerokratt/DMR/issues/14


apps/dmr-server/src/modules/centops/centops.service.ts
```diff
@@ -100,10 +100,20 @@ export class CentOpsService implements OnModuleInit {
...
+
+    return base64;
+  }
+
   async syncConfiguration(): Promise<ClientConfigDto[] | undefined> {
     try {
       const { data } = await firstValueFrom(
-        this.httpService.get<IGetAgentConfigListResponse>(this.centOpsConfig.url),
+        this.httpService.get<IGetAgentConfigListResponse>(this.centOpsConfig.url, {
+          params: { pageSize: 100 },
```
IgorKrupenja, line 114: Thanks, this is better for now actually!

IgorKrupenja (approved):

